PR #8, opened 2017-06-25
Add readcuts!

lkapelevich:


src/valuefunctions.jl
```diff
@@ -173,3 +171,5 @@ function rebuildsubproblem!{C<:AbstractCutOracle}(vf::DefaultValueFunction{C}, m
     end
     m.stages[ex.stage].subproblems[ex.markovstate] = sp
 end
+
+rebuildsubproblem!(vf::DefaultValueFunction{DefaultCutOracle}, m::SDDPModel, sp::JuMP.Model) = nothing
```
lkapelevich: @odow  This is back (I can't see a situation where this method would be called), but `loadcuts!` isn't calling `rebuildsubproblem!` anymore. It just calls `addcut!` directly. Was there a reason for going through `rebuildsubproblem!` before?

odow: No. I hadn't even really tested it. This method gets called when cut selection is called with the DefaultCutOracle. I need to refactor things to clear things up

Issue #9, opened 2017-06-25
Typo in examples


lkapelevich:
In the hydro valley examples, generation only includes the last level instead of all levels. E.g.:
https://github.com/odow/SDDP.jl/blob/master/examples/visualisation.jl#L84

odow:
PR?

PR #12, opened 2017-06-27
Fix hydro examples typo

lkapelevich:
Fix https://github.com/odow/SDDP.jl/issues/9


odow:
This must have existed in the other library as well because I just copy-and-pasted them across!

PR #14, opened 2017-07-10
Add iteration keyword

lkapelevich:
This is too get the iteration number into a solvehook, not super essential.
Using length because before iteration 1 there is no log.


odow:
What about coming up with a more robust way to differentiate forward and backward solves?

Something like 
```julia
# replace solve(sp) with
function JuMPsolve(T, sp, m)
    solve(sp)
end
# then you could overload
function JuMPsolve(::Type{ForwardPass}, sp, m)
    solve(sp, ignore_solvehook=true)
end
function JuMPsolve(::Type{BackwardPass}, sp, m)
    solve(sp, require_duals=true, iteration=length(m.log))
end
```

PR #25, opened 2017-09-06
Add stagewise noise to objective function

odow:
@adow031 this should work. Not overly tested for bugs.


odow:
Seems to work so merging this now. Will add more examples and tests later.

Issue #6, opened 2017-05-30
Numerical Issues


odow:
Occasionally with
 - Gurobi I get `status: Numeric` 
 - CPLEX I get `InfOrUnBnd`
 - CPLEX I get `CPX_OPT_INFEAS` [here](https://www.ibm.com/support/knowledgecenter/en/SS9UKU_12.5.0/com.ibm.cplex.zos.help/refcallablelibrary/html/macros/CPX_STAT_OPTIMAL_INFEAS.html)
 - Clp I get crashes, or incorrect solutions. 
 - Non-repeatable "Infeasible" status using Gurobi v7.0
 - Non-repeatable "Unbounded" status using Gurobi v7.0

Action items:
- Isolate cause
- Figure out how to make subproblems more numerically robust

This will be significantly helped by the JuMP re-write.

odow:
I've tracked down a source of instability in Gurobi but don't have a MWE yet. Basically, 

Given a state variable `X >=0` we define an incoming state variable `X0 free`.  There is a linking constraint `X <= X0`.

At some point, we get the solution `X(t) = 0.0`. We query this via JuMP and get something like `-5.5e-8` which is smaller than the [default tolerance](http://www.gurobi.com/documentation/7.5/refman/feasibilitytol.html#parameter:FeasibilityTol) of 1e-6. However, when we set this as the value of the incoming state variable in the next stage we get
```
X >= 0.0
X0 == -5.5e-8
X <= X0
```
and Gurobi reports infeasible.

odow:
Leaving this here as a reminder. I attended a webinar run by Gurobi about numerical issues. It turns out that their feasibility tolerance of `1e-6` is absolute rather than relative. If you have a constraint with expressions that evaluate to high orders of magnitude (say billions of dollars), the LHS is compared to the RHS at 1e-6 (i.e. across 15 orders of magnitude). They recommend keeping expressions below `1e4`. Presumably other solvers are the same.

PR #72, opened 2017-11-09
Improve numerical stability

odow:
I finally tracked down a long standing numerical issue in SDDP.jl (ref https://github.com/odow/SDDP.jl/issues/6#issuecomment-343022931).

A state variable can be slightly infeasible in the optimal solution of one subproblem and satisfy the feasibility tolerances. However when set as the incoming state variable in the next subproblem, it can cause the full problem to be infeasible. 

The fix is to project the variable value back into the proper bounds. 

cc @leclere @blegat @frapac @joaquimg you probably want to implement this?


blegat:
Thanks for the warning, that's indeed a good idea. Which solver returns a slightly infeasible solution ? Projecting to bounds will fix bound infeasibility but do you also observe constraint infeasibility ?

odow:
Gurobi. 

> Projecting to bounds will fix bound infeasibility but do you also observe constraint infeasibility ?

We typically assume relatively complete recourse so no. In your situation with feasibility cuts, I guess there is the potential. That would be much harder to fix.

joaquimg:
You were using barrier method or simplex? or default?

odow:
All defaults.

PR #74, opened 2017-11-29
Add dro

lkapelevich:
Add DRO code


odow:
Can the hydro valley example use https://github.com/odow/SDDP.jl/blob/master/examples/HydroValleys/hydro_valley.jl#L45?

test/runtests.jl
```diff
@@ -23,5 +23,6 @@ include("noises.jl")
 include("cutoracles.jl")
 include("stageobjectives.jl")
 include("sddpmodels.jl")
+include("dro.jl")
```
odow: s/dro/DRO

src/DRO.jl
```diff
@@ -0,0 +1,122 @@
...
+        original_distribution::Vector{Float64},
+        observations::Vector{Float64},
+        m::SDDPModel,
+        sp::JuMP.Model
+    )
+
+     # Number of noises
+    S = length(observations)
+
+    updateprobabilities!(newprobabilities, measure, getsense(sp), observations, S)
```
odow: why do we have this function separated into two different ones?

src/DRO.jl
```diff
@@ -0,0 +1,122 @@
...
+type DRO <: SDDP.AbstractRiskMeasure
+    radius::Float64
+end
+
+function popvar(x::Vector{Float64})::Float64
+    ninv = 1 / length(x)
+    ninv * sum(x.^2) - (ninv * sum(x))^2
+end
+
+function popstd(x::Vector{Float64})
```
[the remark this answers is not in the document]
lkapelevich: we didn't like the keyword argument a while ago

src/DRO.jl
```diff
@@ -0,0 +1,122 @@
...
+#  This Source Code Form is subject to the terms of the Mozilla Public
+#  License, v. 2.0. If a copy of the MPL was not distributed with this
+#  file, You can obtain one at http://mozilla.org/MPL/2.0/.
+#############################################################################
+
+export DRO
+
+#==
+This file relates to distributionally robust SDDP, see:
+http://www.epoc.org.nz/publications.html
```
odow: Can you give a better reference? i.e. as a tech report for EPOC

src/DRO.jl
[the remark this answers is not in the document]
odow: Haha okay. Can you add a comment to remind us please. Julia v0.7 should remove the overhead of using keywords so we can change it then.

src/DRO.jl
```diff
@@ -0,0 +1,122 @@
...
+        original_distribution::Vector{Float64},
+        observations::Vector{Float64},
+        m::SDDPModel,
+        sp::JuMP.Model
+    )
+
+     # Number of noises
+    S = length(observations)
+
+    modifyprobability!(newprobabilities, measure, getsense(sp), observations, S)
```
odow: oops. I meant inlining this function rather than having another function call.

lkapelevich: The reason it's been moved out is to separate the maths and make it easier to test, avoiding having to pass in an SDDP model and a JuMP model to the function. I can inline it if you prefer, but then the tests will need have some empty models.

lkapelevich:
I don't have permission to merge this PR right now

lkapelevich:
Could you approve/merge?

src/DRO.jl
odow: That's a good enough reason for me.

Issue #84, opened 2018-02-02
Rename ':noise' key in solution storage


vitorlppf:
Greetings,

I'm working on an out-of-sample simulation function based on the existing 'simulate' functions. Currently it receives a list of SDDP.Noise objects (of size equal to the number of stages). I created methods to save the results, including the values of the SDDP.Noise argument in a key temporarily named ':new_noise'.

I think the name ':noise' for the storage key that represents the index for the noise in the model is unclear. Especially if you intend to implement sampling oracles as per #80. I'd like to suggest renaming the ':noise' key to ':noiseidx' or similar. What's your opinion on this?

Thanks,
Vitor.

odow:
Hey Vitor, implementing an out of sample simulator sounds great! 

One issue is that renaming the key would break a lot of existing code. 

I'd be more in favour a bigger reorganisation that included returning the actual noise object (or implementing the oracles since they seem the next most important thing).

vitorlppf:
I renamed the keys in my local branch, and it broke my own code! It was a trivial fix, but I can see how it would disrupt other people's code. I believe a way of implementing this change without such issues would be creating a new `simulate` method, and deprecating the existing ones.

I think having `0` indicate when `:noiseidx` is sensible, because the output would still always be a (non-negative) `Int`. It's easy to program around too. I was thinking about `:noise` being an array with the noise values, but associating it with [`Noise`](https://github.com/odow/SDDP.jl/blob/master/src/typedefinitions.jl#L63-L70) objects sounds like a better idea.

I could implement the key name changes and `SamplingOracles` (#80) as well, if you haven't worked on it already. Since you probably already gave it some thought, is there any particular detail you have in mind for them?

odow:
I haven't worked on them. They would be a great contribution!

The only thoughts I've had I put down in #80. They seem like a good starting point.

I think once we have sampling oracles, figuring out a new (improved) simulation method will be easier.

> I was thinking about :noise being an array with the noise values

This gets a little tricky because I often go
```julia
w = [1, 2, 4]
@rhsnoise(sp, i=1:3, x <= w[i])
```
Should the noise values be `1,2,3` or `1,2,4`?

I would suggest making a Pull Request that begins to implement sampling oracles, and we can continue the discussion there. I'm sure we will run into a few problems.

Issue #80, opened 2018-01-30
Proposal: Sampling oracles


odow:
Implement an oracle that can be used to sample the [subproblem](https://github.com/odow/SDDP.jl/blob/7f1e47140efc4fd33f0cca2be27c0331bef11203/src/SDDP.jl#L215) or [noise](https://github.com/odow/SDDP.jl/blob/7f1e47140efc4fd33f0cca2be27c0331bef11203/src/SDDP.jl#L222) in the forward pass of the algorithm. 

It probably takes the form of a new overloadable type that gets dispatched in the above functions. 

Bonus points for solving https://github.com/odow/SDDP.jl/issues/10 in addition to closing https://github.com/odow/SDDP.jl/issues/29 and #79.

Potentially looks something like

```julia
abstract type SamplingOracle end

struct Default <: SamplingOracle end
function samplenoise(::Default, m::SDDPModel, sp::JuMP.Model)
    ex = ext(sp)
    noiseidx = sample(ex.noiseprobability)
    return ex.noises[noiseidx]
end

struct Uniform <: SamplingOracle end
function samplenoise(::Uniform, m::SDDPModel, sp::JuMP.Model)
    return rand(ext(sp).noises)
end

struct Deterministic{T} <: SamplingOracle 
    data::Vector{T}
end
function samplenoise(oracle::Deterministic{T}, m::SDDPModel, sp::JuMP.Model) where T
    oracle.data[ext(sp).stage]
end

# ...

solve(m, sampling_oracle=Default())
```

vitorlppf:
What should be the output of an oracle? Just a `Noise` object, a `Noise` object and an `Int`, or should both be possible?

In #88 I assumed the first, but consequently I'm always saving `0` in place of the index. I pondered if I should change it before committing, but decided to leave it as is.

PR #88, opened 2018-02-11
Sampling Oracles

vitorlppf:
An implementation of Sampling Oracles as suggested in #80, as well as renaming the `:noise` key to `:noiseidx` as suggested in #84.

I apologize for the delay. I wrote down most of the implementation a while ago, but I decided to wait some time before testing to do it with a clear mind. I ran into some minor issues, but I believe the current version should work. I tested it with [simplified_hydrothermal_dispatch.jl](https://github.com/odow/SDDP.jl/blob/master/examples/HydroValleys/simplified_hydrothermal_dispatch.jl).

### DIFF
--- a/src/SDDP.jl
+++ b/src/SDDP.jl
@@ -219,7 +219,8 @@ function forwardpass!(m::SDDPModel, settings::Settings, solutionstore=nothing)
 
         # choose and set RHS noise
         if hasnoises(sp)
-            (noiseidx, noise) = samplenoise(sp, solutionstore)
+            noise = samplenoise(settings.oracle, sp, solutionstore)
+            noiseidx = 0
             setnoise!(sp, noise)
         end
         # solve subproblem
@@ -281,7 +282,6 @@ function iteration!(m::SDDPModel, settings::Settings)
         end
     end
     return objective_bound, time_backwards, simulation_objective, time_forwards
-
 end
 
 function rebuild!(m::SDDPModel)
@@ -434,6 +434,8 @@ control the solution process.
     Defaults to `false`.
  * `cut_output_file::String`:
     Relative filename to write discovered cuts to disk. Defaults to `""` (no cuts written)
+ * `oracle::SamplingOracle`
+    Scenario sampling method. Defaults to `DefaultSamplingOracle()`.
 
 # Returns
  * `status::Symbol`:
@@ -470,7 +472,8 @@ function JuMP.solve(m::SDDPModel;
         # this reduces memory but you shouldn't use it if you want to save the
         # sddp model since it throws away some information
         reduce_memory_footprint      = false,
-        cut_output_file::String      = ""
+        cut_output_file::String      = "",
+        oracle::SamplingOracle       = DefaultSamplingOracle()
     )
     reset_timer!(TIMER)
 
@@ -492,7 +495,8 @@ function JuMP.solve(m::SDDPModel;
         log_file,
         reduce_memory_footprint,
         cut_output_file_handle,
-        isa(solve_type, Asyncronous)
+        isa(solve_type, Asyncronous),
+        oracle
     )
 
     print(printheader, settings, m, solve_type)
--- a/src/noises.jl
+++ b/src/noises.jl
@@ -19,8 +19,8 @@ function samplenoise(sp::JuMP.Model)
 end
 samplenoise(sp::JuMP.Model, solutionstore::Void) = samplenoise(sp)
 function samplenoise(sp::JuMP.Model, solutionstore::Dict{Symbol, Any})
-    if length(solutionstore[:noise])>=ext(sp).stage
-        idx = solutionstore[:noise][ext(sp).stage]
+    if length(solutionstore[:noiseidx])>=ext(sp).stage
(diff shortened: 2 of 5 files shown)


odow:
Awesome. Delays are fine! You are a volunteer. I'll have a better look later, but could you look into overloading `samplesubproblem` as well?

vitorlppf:
I can certainly implement the current `samplesubproblem` method as the Default option for a subproblem sampling oracle. I think it would be better if those were of a different type from `SamplingOracle`. Do you agree? If so, I'm going to rename all the types added in the current implementation, for the sake of clarity.

src/SDDP.jl
```diff
@@ -219,7 +219,8 @@ function forwardpass!(m::SDDPModel, settings::Settings, solutionstore=nothing)
 
         # choose and set RHS noise
         if hasnoises(sp)
-            (noiseidx, noise) = samplenoise(sp, solutionstore)
+            noise = samplenoise(settings.oracle, sp, solutionstore)
+            noiseidx = 0
             setnoise!(sp, noise)
```
odow: I think `samplenoise` should still return the tuple `(noiseidx, noise)`. Users should return the index if it exists in the model, or the index `0` if it is out-of-sample.

src/noises.jl
```diff
@@ -13,20 +13,35 @@ function setnoise!(sp::JuMP.Model, noise::Noise)
...
-    noiseidx = sample(ext(sp).noiseprobability)
-    return noiseidx, ext(sp).noises[noiseidx]
-end
-samplenoise(sp::JuMP.Model, solutionstore::Void) = samplenoise(sp)
-function samplenoise(sp::JuMP.Model, solutionstore::Dict{Symbol, Any})
-    if length(solutionstore[:noise])>=ext(sp).stage
-        idx = solutionstore[:noise][ext(sp).stage]
+function samplenoise(oracle::SamplingOracle, sp::JuMP.Model, solutionstore::Dict{Symbol, Any})
+    if length(solutionstore[:noiseidx])>=ext(sp).stage
+        idx = solutionstore[:noiseidx][ext(sp).stage]
```
odow: This should probably go since the historical simulations can just use a DeterministicOracle.

src/simulate.jl
```diff
@@ -77,14 +77,15 @@ markov state to sample in each stage.
 function simulate(m::SDDPModel,
```
odow: This method should probably be replaced by one that uses a DeterministicOracle.

src/typedefinitions.jl
```diff
@@ -231,6 +233,37 @@ Defaults to `false`.
 """
 MonteCarloSimulation(;frequency=0,min=20,max=0,step=1,confidence=0.95,termination=false) = MonteCarloSimulation(frequency,collect(min:step:max),confidence,termination)
 
+struct DefaultSamplingOracle <: SamplingOracle end
+
+struct UniformSamplingOracle <: SamplingOracle end
+
+struct DeterministicSamplingOracle{T} <: SamplingOracle 
+    data::Vector{T}
+end
```
odow: Can you add doc strings for these oracles like you have for CustomSamplingOracle?

src/utilities.jl
```diff
@@ -156,7 +156,7 @@ end
 samplesubproblem(stage::Stage, last_markov_state::Int, solutionstore::Void) = samplesubproblem(stage, last_markov_state)
 
 function samplesubproblem(stage::Stage, last_markov_state, solutionstore::Dict{Symbol, Any})
-    if length(solutionstore[:noise]) >= stage.t
+    if length(solutionstore[:noiseidx]) >= stage.t
```
odow: See my comment regarding using a DeterministicOracle for historical simulations.

src/noises.jl
vitorlppf: Could you be more clear? Currently I intend to keep the existing methods under `DefaultSamplingOracle`, while having a separate method for `DeterministicSamplingOracle`.

src/utilities.jl
vitorlppf: When do you expect `length(solutionstore[:noiseidx])>=stage.t` to return `true` and when do you expect it to return `false`? It seems to me that the number of indexes stored would always be strictly lesser than the current stage during simulation, as I'd start at stage `1` with a 0-length index array and both would increment simultaneously inside [`forwardpass!`](https://github.com/odow/SDDP.jl/blob/master/src/SDDP.jl#L213).

src/noises.jl
odow: The `solutionstore` was just a hack for running historical simulations. Now we could go
```julia
# Monte-Carlo
simulate(
    m, 
    100,
    oracle = DefaultOracle()
)

# Historical
simulate(
    m, 
    oracle = DeterministicOracle(
        markov_state = [1,1,2],
        noise_index = [1, 2, 1]
    )
)
```

So this whole method can probably be deleted.

src/utilities.jl
odow: This relates to https://github.com/odow/SDDP.jl/pull/88/files#r170729677 

Once `samplesubproblem` is implemented for the oralces, this method can be removed.

odow:
p.s. can I ask where you are based any how you found out about the library?

Issue #92, opened 2018-02-21
Proposal: More Exotic Policy Graphs


odow:
We can elaborate on this once my thesis is written, saving this here while I remember.

This is a nice way to avoid confusing the whole stage/node/`do sp, t`/`do sp, t, i` situation.
```julia
struct PolicyGraph{T}
    root::T
    subproblems::Dict{T, Vector{Tuple{T, Float64}}}
end

function LinearGraph(stages::Int)
    subproblems = Dict{Int, Vector{Tuple{Int, Float64}}}()
    subproblems[0] = Tuple{Int, Float64}[]
    for t in 1:stages
        subproblems[t] = Tuple{Int, Float64}[]
        push!(subproblems[t-1], (t, 1.0))
    end
    PolicyGraph(0, subproblems)
end

function MarkovianGraph(transition_probabilities::AbstractVector{AbstractArray{Float64, 2}})
    # check is Markov transition matrix
    i = 1
    for T in transition_probabilities
        @assert size(T, 1) == i
        i = size(T, 2)
        for j in 1:size(T, 1)
            @assert isapprox(sum(T[j,:]), 1.0)
        end
    end

    typ = Tuple{Int, Int}
    subproblems = Dict{typ, Vector{Tuple{typ, Float64}}}()
    subproblems[(0,1)] = Tuple{typ, Float64}[]
    for (t, T) in enumerate(transition_probabilities)
        for j in 1:size(T, 2)
            subproblems[(t, j)] = Tuple{typ, Float64}[]
        end
        for i in 1:size(T, 1)
            for j in 1:size(T, 2)
                push!(subproblems[(t-1, i)], ((t, j), T[i,j]))
            end
        end
    end
    PolicyGraph((0,1), subproblems)
end
```

```julia
SDDPModel(
    policy_graph = LinearGraph(3)
) do sp, index
    stage = index
    # stage = 1,2,3
end

SDDPModel(
    policy_graph = MarkovianGraph([ 
        [1.0]',
        [0.2 0.8],
        [0.5 0.5; 0.1 0.9]
    ])
) do sp, index
    (stage, markov_state) = index
    # (stage, markov_state) = (1,1), (2,1), (2,2), (3,1), (3,2)

end
```

```julia
function children(g::PolicyGraph{T}, subproblem::T) where T
    [s[2] for s in g.subproblems[subproblem]]
end

g = LinearGraph(3)
children(g, 0) # [1]
children(g, 2) # [3]
[...]

PR #94, opened 2018-02-27
Discount prices over time by interest rate

bfpc:
It is usual, in long-term planning, to apply a (usually fixed) interest rate to the future cost.

I expect this small change does everything it's supposed to, but I'm not 100% sure.  I was wondering if this should create an entirely new discountedValueFunction.jl, but I don't see the point.  Maybe, for changing interest rates over time this could be justified...

I have tested against examples/simple_objective_noise.jl , examples/Newsvendor/newsvendor.jl and examples/HydroValleys/hydro_valley_tests.jl

### DIFF
--- a/src/defaultvaluefunction.jl
+++ b/src/defaultvaluefunction.jl
@@ -11,6 +11,7 @@ getstageobjective(sp::JuMP.Model) = getstageobjective(valueoracle(sp), sp)
 mutable struct DefaultValueFunction{C<:AbstractCutOracle} <: AbstractValueFunction
     cutmanager::C
     theta::JuMP.Variable
+    discount_rate::Float64
 end
 
 """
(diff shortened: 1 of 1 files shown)


odow:
I've implemented discounting in the past by multiplying the stage objective by the discount factor:
```julia
SDDPModel() do sp, t
    ρ = 0.95
    @stageobjective(sp, ρ^(t-1) * ( ... normal objective ... ) )
end
```
Does this achieve the same thing?

odow:
As an aside, you can run the packages unit tests by opening a new Julia REPL and running `Pkg.test("SDDP")`. This will run all the tests in the [`/test`](https://github.com/odow/SDDP.jl/tree/master/test) folder. If you run the tests (or look at the Travis log), you'll see that [this test](https://github.com/odow/SDDP.jl/blob/master/test/stageobjectives.jl#L17) is now failing because JuMP must reorder the variables now that the coefficients have changed.

It would be good to add a new example showing how to introduce a discount rate and confirm that it converges to the discounted lower bound.

bfpc:
The explicit exponential term should indeed yield the same result - it's just that it seems more natural to discount at every stage.

Should I correct the failing test, or is it better to encourage incorporating the discount at each `@stageobjective`?

odow:
The general rule I have followed is, where possible, to avoid incorporating special features than can be implemented by the user (for example, AR(N) processes for rainfall). 

I have a much better way to implement this that naturally extends to infinite horizon models with discount factors, however I'm not planning on implementing it until after I submit my PhD (~May). Unfortunately it is linked to #92 and will require a significant re-write of the library.

In the meantime, I would gladly accept a PR for a new example with discounting that solves a couple of times for different discount factors showing the different results.

odow:
Closing in favour of #138. This just needs some documentation until we can redo this better.

PR #125, opened 2018-04-29
Allow users to save dual variables during simulation

bfpc:
It can be interesting to save dual variables to get marginal costs for some resources.

### DIFF
--- a/examples/HydroValleys/simplified_hydrothermal_dispatch.jl
+++ b/examples/HydroValleys/simplified_hydrothermal_dispatch.jl
@@ -47,7 +47,7 @@ m = SDDPModel(
     # ------------------------------------------------------------------
     #   Constraints
     # Demand
-    @constraint(sp, g[1] + g[2] + u == 150)
+    @constraint(sp, demand_eq, g[1] + g[2] + u == 150)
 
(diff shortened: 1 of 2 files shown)


odow:
Can you add an example that tests this please?

I think you really want to get the dual on the incoming state variables. 
https://github.com/odow/SDDP.jl/blob/338030afd74ade9ed8b68eabaa4afb179694908d/src/states.jl#L12

bfpc:
Do you mean something for code coverage, or a "toy problem"?  I can try to write a simple example for this one, but I don't know of any exactly solved (3+)-stage problem for which the derivatives are known.

Thanks for the pointer for the subgradient wrt the state variables.  However, what motivated me in this case was the need for the marginal cost for the demand, which is not a state variable, but a RHS of a balance equation.

odow:
You could just modify this example
https://github.com/odow/SDDP.jl/blob/338030afd74ade9ed8b68eabaa4afb179694908d/examples/HydroValleys/simplified_hydrothermal_dispatch.jl#L50

I just want something for code coverage so that when the Travis tests run the line gets hit.

odow:
Thanks @bfpc !

PR #139, opened 2018-06-09
Allows users to savel all/only valid cuts

bfpc:


odow:
Thanks @bfpc. I was just thinking about this yesterday! I think we need to add a function
```julia
"""
    allcuts(oracle)

Return all of the cuts stored in the oracle, not just those returned 
by `validcuts(oracle)`.
"""
function allcuts end
```
to [this file](https://github.com/odow/SDDP.jl/blob/master/src/cut_oracles/cutoracles.jl). Then we can call `allcuts(cut_or)` instead of relying on `cut_or.cuts`.

If you're busy, I can do this later today.

odow:
Thanks @bfpc.

I fixed a couple of mistakes and added some docs and tests.

Issue #36, opened 2017-09-11
Tidy up d3 visualiser


odow:
At the moment it's horrible (because I know nothing about JS).

https://github.com/odow/SDDP.jl/blob/master/src/visualizer/simulation.js

Bonus points for making it easy to embed into other sites such as @EyobZewdie's Part IV report.

PR #146, opened 2018-06-16
Fix plotting on *NIX machines

### LINKED ISSUE #145: Minor problem with graphing
I'm using your (great so far!) library and following your tutorials. I would like to report a minor issue: when graphing the value functions, the generated file using plotly is a temp file with no html extension (at least on Linux). Hence the browser window pops up but it does not parse the html output.

I tracked the issue to `launch_file` in `visualize.jl` where you do:
`launch_file(html_string, assets, outputfile=replace(tempname(), ".tmp", ".html"))` 

However tempname() does not return an extension .tmp, so nothing gets replaced. It should be:
`launch_file(html_string, assets, outputfile=tempname() * ".html")`

Thanks!
Andrés

odow:
Closes #145

### DIFF
--- a/src/visualizer/visualize.jl
+++ b/src/visualizer/visualize.jl
@@ -8,7 +8,7 @@
 
 const ASSET_DIR = dirname(@__FILE__)
 
-function launch_file(html_string, assets, outputfile=replace(tempname(), ".tmp", ".html"))
+function launch_file(html_string, assets, outputfile)
     for asset in assets
         cp(joinpath(ASSET_DIR, asset), joinpath(dirname(outputfile), asset), remove_destination=true)
     end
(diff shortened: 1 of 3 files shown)


odow:
Hi @aferragu do you mind testing that this works? 

You can either checkout out this PR using Julia:
```julia
Pkg.checkout("SDDP", "fix-visualize")
```
Then, once this branch is merged you can get back to master by:
```julia
Pkg.free("SDDP")
# or 
Pkg.checkout("SDDP", "master")
```

Alternatively, if you know much about Git you can just do it via the command line.

aferragu:
That should do it. I can confirm it works on *NIX. Thanks!

odow:
Thanks @aferragu! If you know anything about JavaScript, #36 needs work :)

PR #148, opened 2018-06-21
Add experimental (hacky) support for noise in the constraint matrix.

odow:
Developed after hacking with @Thuener 

Largely un-tested and under-development.


Thuener:
Thanks @odow. It is working fine.

odow:
Awesome. Interesting to hear how it goes and if we can improve it

Issue #120, opened 2018-04-17
Markov State behavior


rodripor:
Hello,

When running SDDP algortihm with Markov State model in all forward passes the first stage state is 1. Is there a way to modify this behavior through a parameter? 

When the model is resolved, i can plot valuefunction depending on markov state but i cant get explicitly the set of cuts associated with that state in that stage. Is there a way to obtain such cuts from framework? I didnt find a structure where this cuts are stored.

Thanks for your help,
Rodrigo

odow:
> Is there a way to modify this behavior through a parameter?
The `markov_transition` argument to `SDDPModel` takes a `Vector{Array{Float64, 2}}` with one element for each stage. Each element is a `MxN` matrix. We read the matrix `T[i,j]` as the probability of transitioning from Makov state `i` in stage `t-1` to Markov state `j` in stage `t`. The first stage is special as it assumes there is a "zero-th" stage (which I will call the root node) with one Markov state. 

If you want two Markov states in the first stage, then the first element is a `1x2` array. 

For example, here is the transition matrix for a problem with two stages, and two Markov states within each stage. 
```julia
transition = Array{Float64, 2}[
    [ 0.5 0.5 ],
    [ 0.6 0.4 ; 0.4 0.6]
]
```

> Is there a way to obtain such cuts from framework?

You have two options. The first programatically returns a list of [`SDDP.Cut` objects](https://github.com/odow/SDDP.jl/blob/7f1e47140efc4fd33f0cca2be27c0331bef11203/src/typedefinitions.jl#L53-L56)
```julia
STAGE, MARKOV   = 1, 2
sp = SDDP.getsubproblem(m, STAGE, MARKOV)
oracle = SDDP.cutoracle(sp)
cuts = SDDP.validcuts(oracle)
```

The second is to use the `cut_output_file` argument to solve. 
```julia
solve(m, cut_output_file = "mycuts.csv")
```
This will produce a CSV file containing a list of all the cuts. The columns are
`STAGE, MARKOV STATE, intercept, state 1 coefficient, state 2 coefficient ...`

odow:
Also, if you have some examples with multiple Markov states in the first stage, it would be great to include them in the library. I'm always looking for new models.

rodripor:
> Also, if you have some examples with multiple Markov states in the first stage, it would be great to include them in the library. I'm always looking for new models.

[...]

PR #152, opened 2018-07-19
Fix bound calculation with multiple Markov states in first stage

odow:
As reported by @rodripor, the bound calculation did not account for multiple Markov states in the first stage.

There is an unresolved question as to what the bound should be when there are multiple Markov states (or stagewise-independent noise terms for that matter). Should it just be the expectation of all of the possible first stage realizations? Or some risk-adjusted value? Even though we do not explicitly optimize the risk-adjusted value, it should attain the optimal solution from monotonicity.

Closes https://github.com/odow/SDDP.jl/issues/120

### DIFF
--- a/src/SDDP.jl
+++ b/src/SDDP.jl
@@ -245,11 +245,27 @@ function backwardpass!(m::SDDPModel, settings::Settings)
             end
         end
     end
+    return compute_initial_bound(m)
+end
+
+"""
+    compute_initial_bound(m)
+
+Calculate the expected cost of the first stage.
+"""
+function compute_initial_bound(m)
+    # TODO(odow): should this return the risk-adjusted expectation? This implies
+    # that we need a risk measure at the root node.
     reset!(m.storage)
     for sp in subproblems(m, 1)
         solvesubproblem!(BackwardPass, m, sp)
     end
-    return dot(m.storage.objective, m.storage.probability)
+    transition_matrix = getstage(m, 1).transitionprobabilities
+    for (i, markov_state) in enumerate(m.storage.markov)
+        m.storage.probability[i] *= transition_matrix[1, markov_state]
+    end
+    @assert sum(m.storage.probability) ≈ 1.0
+    dot(m.storage.objective, m.storage.probability)
 end
 
 function iteration!(m::SDDPModel, settings::Settings)
--- a/test/sddpmodels.jl
+++ b/test/sddpmodels.jl
@@ -136,3 +136,14 @@ using Clp
         end
     end
 end
+
+@testset "Markov Issue #120" begin
+    m = SDDPModel(stages=2, sense=:Min, objective_bound=0.0, solver=ClpSolver(),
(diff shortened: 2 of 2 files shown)


odow:
Pinging @rodripor, are you using `Expectation()` as your risk measure? What do you expect the bound to be if there are multiple Markov states in the first stage?

aferragu:
Let me pitch in here Oscar (I'm @rodripor advisor). We are using `Expectation()` as risk measure. However, seems to me that it doesn't make sense to have multiple Markov states for t=0 since the cost-to-go is a function of the state, both its continuous part and its Markov part. Therefore, the bound should be for a given initial Markov state and continuous states. 

There is some asymmetry in your definition of continuous and Markov states. Whenever you define a state variable with the  `@state` macro, you give an initial condition (the third argument). However, for the Markov transitions you expect to have a probability fixed to 1 to one of them, and then a vector for the first jumps, and finally the transition matrices. Maybe a better way to go is to define the transition matrices, and have a Markov state constructor where you can specify the initial condition. I'm not sure if your code can be refactored to perform this, due to the way you define Markov transitions.

Regarding your proposed fix, I don't understand this line:
`m.storage.probability[i] *= transition_matrix[1, markov_state]`
unless `m.storage.probability[i]` is initialized to 1. However, I don't know if you need this fix or impose an initial Markov state instead.

odow:
> Maybe a better way to go is to define the transition matrices, and have a Markov state constructor where you can specify the initial condition. I'm not sure if your code can be refactored to perform this, due to the way you define Markov transitions.

In [Chapter 1](https://odow.github.io/SDDP.jl/latest/assets/dowson_thesis.pdf) of my thesis, I introduce the idea of a _policy graph_ as a way of structuring stochastic optimization problems. Basically, instead of thinking about stages and Markov states, we should think of nodes in some (potentially cyclic) graph. 

This leads to two possibilities:
1. In my thesis I define a _root node_. When viewed this way, it is fine to have multiple Markov states in the first "stage". Effectively you find a decision rule for each node _a priori_.
2. However, I have been thinking about changing this, and defining the initial conditions as a `(node, noise realization, initial state)` triplet. With this definition, the first "stage" problem is deterministic, and all the uncertainty happens in future stages. 

Option (1) more of less corresponds to what we have at the moment. Option (2) more of less corresponds to what you are proposing.

Refactoring into policy graphs is a major re-write, so I have been putting it off. Introducing a parameter for the initial Markov state should be relatively straight forward.

> Regarding your proposed fix, I don't understand this line:
`m.storage.probability[i] *= transition_matrix[1, markov_state]`

This is a little sleight of hand. The code isn't clear and needs to be re-written. After solving all of the problems in a stage, `m.storage.probability` ends up containing the probabilities of the noise outcomes ignoring the Markov transition. So for two Markov states, each with two noise realizations, `m.storage.probability = [0.5, 0.5, 0.5, 0.5]`. Therefore, we need to weight by the Markov transitions.

odow:
I'm merging this as-is. I'm about to embark on a re-write that will address a lot of these questions.

PR #155, opened 2018-07-23
Add a distributionally-robust risk measure based on the Wasserstein distance

odow:
@dukduque this is based on Dave's talk at ISMP. 

Instead of bundling everything up into the extended subproblem, we just solve a <s>QCP</s>LP on the side to work out what the worst-case probability is.

<s>Did I implement the QCP correctly?</s>No.


dukduque:
Hey! So I saw the model, but I think that is DRO with Phi divergence (or maybe I looked the wrong file). If x are decision variables for the probabilities, we still need the transportation variables (to transport mas from the original distribution to 'x'. Also, I couldn't tell how the support is used.

odow:
Oops, good point. I couldn't really remember the formulation so I just made something up. Here is, hopefully, something more correct.

dukduque:
I think there is still something to fix in the model. So if observations is being used for both the abs and the objective.  If observations are the support points of the random variables, is correct in the abs and incorrect in the objective. The coefficients of the "adjusted_probability" (I'll call this p) should be the objective value of the corresponding (descendant) subproblems. If it is the other way arround, the objective is find and the abs should be reassuring the distance between the support points.

So at iteration t, if q is the nominal distribution of the outcomes of stage t+1 and z is the transportation variables, the at iteration k is:
min \sum_{j=1..N} \cQ_{j,k}*p_j
s.t.
sum_{j=1..N}z_ij = q_i  \forall i=1..N
sum_{i=1..N}z_ij = p_j  \forall j=1..N
sum_{i=1..N, j=1..N} ||\xi_i  - \xi_j || z_ij <= \alpha.  #DRO CONSTRAINT
z>=0

with  \cQ_{j,k} being the objective value of subproblem j (j \in {1,...N}) at stage t+1 at iteration k (i.e., \cQ is the approximation of the value function Q).

odow:
Argh. I'm confused. What is `xi`? The random variable? I think for our case, `\xi_i = Q_{i,k}`, hence the confusion.

- `observations` are the approximations of the value function for the next stage for realizations `i=1,...,N` (parameter)
- `original_probability` is the nominal probability distribution (parameter)
- `adjusted_distribution` is the risk-adjusted probability distribution (variable)
- `transport` is the transport matrix (variable)

PR #180, opened 2019-03-13
Merge kokako

### LINKED ISSUE #6: Numerical Issues

### LINKED ISSUE #10: A better way to run simulations using different noises to the original model
A common pattern is to define a set of noises, converge the model, and then simulate using some out-of-sample observations.

The current approach is to dump the cuts to file when solving the model, then rebuild a different model with the new observations, load the cuts and simulate.

This should be improved.

Should be resolved along with #29

### LINKED ISSUE #28: Remove the need to specify `objective_bound`
To do this, begin with random forward pass that ignores the unboundedness to avoid specifying an initial bound. 

Bonus points for tying into #63

### LINKED ISSUE #29: Add ability to sample forward pass with different probabilities
For example, forward sample with the `riskadjusted_distribution` of the risk measure, or uniformly.

### LINKED ISSUE #36: Tidy up d3 visualiser

### LINKED ISSUE #70: Stagewise indepdendent noise in constraint matrix
The requests keep piling up for this one.

Waiting until JuMP/MOI is released.

### LINKED ISSUE #79: Best way to use historical inflow data in dams model
Hi, 

Im developing an example of hydro dams model. I want to use some historical data to generate trajectories extracting series directly from weekly data in a deterministic way. The first trajectory to use gets the first week of history and advances in that serie, the second trajectory gets the second week and so on. If historical data is shorter than trajectory i get modulus and get the initial data again. When trying to develop in this way i cant find a way in the framework to get the trajectory identifier to index the first week that i must take (if first trajectory the initial week must be week 1, if nth trajectory the initial week must be nth % size). Anyone had this issue? Any suggestions to avoid changing framework code? 

Thanks,
Rodrigo

### LINKED ISSUE #80: Proposal: Sampling oracles

### LINKED ISSUE #84: Rename ':noise' key in solution storage

### LINKED ISSUE #92: Proposal: More Exotic Policy Graphs

### LINKED ISSUE #96: State dependent lipschitz constant
The `lipschitz_constant` argument to the dynamic interpolation method could take a tuple giving a different constant for each of the different price dimensions.

As suggested by @adow031

### LINKED ISSUE #107: Proposal: change noise syntax
```julia
@rhsnoise(sp, w=[1,2,3], 2x+y<=w)
# becomes
@rhsnoise(sp, [1,2,3], w->2x+y<=w)
```
This sorts out the using different noise fiasco since we can just throw new observations into the function (ref #10). It plays nicer with namespacing issues
```julia
w = 0
@rhsnoise(sp, w=[1,2,3], 2x+y<=w)
# what is w? 3!

w = 0
@rhsnoise(sp, [1,2,3], w->2x+y<=w)
# what is w? 0!
```

And is probably a good start towards noises in the constraint coefficients. It also slightly unifies our price interpolation stuff since we would have
```julia
@stageobjective(sp, [1,2,3], w->w*x)
# and
@stageobjective(sp, p->p*x)
```
But we should really have
```julia
@stageobjective(sp, [1,2,3], (p, w) -> p*x + w)
```
to allow for stagewise independent noise on-top-of the price states.

Nothing actionable yet, leaving this as a reminder.

### LINKED ISSUE #115: Add a check for numerical stability
It would be useful to go through and check numerical stability in each of the subproblem (#6).

- Analyze coefficients for user-warning. Things to check
  - [ ] Large coefficients > 1e7
        `log10(maximum(abs.(coefficients))) > 7`
  - [ ] Small coefficients < 1e-2
        `-6 < log10(minimum(abs.(coefficients))) < -2`
  - [ ] High dynamic range 
        `log10(maximum(abs.(coefficients)) - minimum(abs.(coefficients))) > 8`
- [ ] Throw away cuts that are numerically almost identical.

We could report something like Gurobi does
```
Optimize a model with 44 rows, 19 columns and 106 nonzeros
Coefficient statistics:
  Matrix range    [2e-05, 1e+03]
  Objective range [1e-04, 9e+02]
  Bounds range    [1e+01, 4e+05]
  RHS range       [2e+00, 6e+05]
```

```
-------------------------------------------------------------------------------
                      SDDP Solver. © Oscar Dowson, 2017.
-------------------------------------------------------------------------------
    Solver:
        Serial solver
    Model:
        Stages:         4
        States:         2
        Subproblems:    7
        Value Function: Default
        Numerical Stability Issues: none
-------------------------------------------------------------------------------
```

```
-------------------------------------------------------------------------------
                      SDDP Solver. © Oscar Dowson, 2017.
-------------------------------------------------------------------------------
    Solver:
        Serial solver
    Model:
        Stages:         4
        States:         2
        Subproblems:    7
        Value Function: Default
        Numerical Stability Issues:
            Objective Coefficients: 
                Stage 1, Markov State 1: [-1e-2, +1e13]
             Constraint Coefficients: 
                Stage 13, Markov State 3: [-1e-2, +1e13]
                Stage 14, Markov State 3: [-1e-2, +1e13]
                Stage 14, Markov State 3: [-1e-2, +1e13]
-------------------------------------------------------------------------------
```

This might also be useful
```julia
function getlinconstrmatrix(m)
    A = zeros(length(m.linconstr), MathProgBase.numvar(m))
    for (row, con) in enumerate(m.linconstr)
        for (coef, var) in zip(con.terms.coeffs, con.terms.vars)
            A[row, var.col] += coef
        end
    end
    A
end
```

### LINKED ISSUE #118: Refactor
- `@state(sp, x_out >= 0, x_in==1)` -> `@state(sp, x_in==1, x_out >= 1)`
- <s>`max_iterations` -> `iteration_limit` to line up with `time_limit`</s> #136
- `simulation = MonteCarloSimulation()` -> `statistical_stopping_rule = SequentialSampling()`
- <s>`bound_stalling = BoundConvergence()` -> `bound_stalling_stopping_rule = BoundConvergence()`</s> #137
- `cut_selection_frequency` -> `cut_selection_period`
or 

```
iteration_limit = 100,
time_limit      = 20,
stopping_rules  = [
    Statistical(
        min       = 0,
        max       = 1_000,
        step      = 100,
        frequency = 10,
        terminate = false
    ),
    BoundStalling(
        atol       = 0.0,
        rtol       = 0.0,
        iterations = 5
    )
]
```

### LINKED ISSUE #124: Proposal: Stopping Rules
something along the lines of
```julia
abstract type StoppingRule end

struct BoundStalling <: StoppingRule
    N::Int
    tol::Float64
end

function stoppingrule(sr::BoundStalling, m::SDDPModel)
    for i in 0:min(length(m.log) - 2, sr.N)
        if abs(m.log[end - i].bound - m.log[end-i-1].bound) > sr.tol
            return false
        end
    end
    return true
end

# usage
solve(m,
    stopping_rules = [
        BoundStalling(5, 0.0)
    ]
)

```

### LINKED ISSUE #149: Document overloadable functions
e.g., https://github.com/odow/SDDP.jl/blob/5c1d31793c6c6d20e8eb94a823c6eb099968a373/src/utilities.jl#L173-L198

@bfpc might have some more.

### LINKED ISSUE #150: Accessing the stage-objective in a simulation
@lorenzoreus writes

> I have a quick question.  I’m using the EAVaR and I  don’t know how to get the cost at each simulation. For example, without using a risk measure, I can get the cost using
```julia
sim = simulate(m, SIMN...)

for i=1:SIMN
        cost[i]=sim[i][:objective][1]
    end
```
> But now `sim[i][:objective][1]` has the AVaR included. I want to get the `cost[i]`. With previous loop, I’m getting `risk_measure(cost[i])`. 
I couldn’t find something on this matter in a thread. May be because is too easy. 

The answer is that there is a `sim[i][:stageobjective][t]` field. So 
```julia
cost = [ sum(s[:stageobjective]) for s in sim ]
```

This needs to be documented better.

Edit: It turns out that it is at the bottom of [Tutorial Two](https://odow.github.io/SDDP.jl/latest/tutorial/02_rhs_noise.html#Tutorial-Two:-RHS-noise). In addition to the tutorials, we probably need a "How do I" section of the docs.

### LINKED ISSUE #156: Get dual variables out of the simulation
It doesn't appear the code is currently set up to allow the easy extraction of dual variables of the subproblem constraints. Could you point me to the place in the code that creates a dictionary reference for a variable 'x' as ':x'?

I'd like to do the same for the constraints.

### LINKED ISSUE #158: Some comments about the package
Hi Oscar:

I've found that when you want to simulate after you have solved the problem. You only can simulate the variables that belong to the same period. I think that this isn't reported in the documentation.

Kind Regards

### LINKED ISSUE #162: Fix Wasserstein Properly
As @dukduque noted, my second attempt was still incorrect. See https://github.com/odow/SDDP.jl/pull/155#issuecomment-420793333.

For third time lucky, we need to cache a vector of the noise coefficients for each subproblem. These can be found:
- right-hand sides
https://github.com/odow/SDDP.jl/blob/0d15d4ab5b9494c4cc96719195cd39885fe29138/src/typedefinitions.jl#L69
- cost coefficients
https://github.com/odow/SDDP.jl/blob/0d15d4ab5b9494c4cc96719195cd39885fe29138/src/typedefinitions.jl#L65
The cost coefficients are going to be a pain as we will need to do a diff on the `AffExpr` objects to work out what they are :(

These vectors of the noise coefficients then become the `xi` terms mentioned in the above linked comment.

### LINKED ISSUE #166: Style Fixes
I'm about to embark on a large re-write of this library. As part of that, I will be updating the code to be much more readable. As such, SDDP.jl will be adopting the [JuMP style guide](http://www.juliaopt.org/JuMP.jl/latest/style.html).

This issue will track the bits I have updated.
- [x] cut oracles https://github.com/odow/SDDP.jl/pull/164
- [x] risk measures https://github.com/odow/SDDP.jl/pull/165
- [ ] price interpolation
- [ ] defaultvaluefunction
- [ ] deprecate
- [ ] experimental
- [ ] JuMPfunctions
- [ ] MIT_licensedcode
- [ ] noises
- [ ] print
- [ ] SDDP.jl
- [ ] simulate
- [ ] solve_asynchronous
- [ ] states
- [ ] typedefinitions
- [ ] utilities
- [ ] `/test`
- [ ] `/examples`

### LINKED ISSUE #172: SDDP.jl, JuMP 0.19, and Julia 1.0
Dear all,

This post is to update you on my plans for the future of SDDP.jl.

Currently `SDDP.jl` supports Julia v0.6 and JuMP v0.18. JuMP v0.18 has recently added support for Julia v1.0. However, because of the way we implement macros like `@state` and `@rhsnoise`, upgrading `SDDP.jl` to Julia v1.0 is non-trivial. (This is mainly due to some changes in the way scoping works in macros.)

In addition, soon, [JuMP will be releasing v0.19](https://discourse.julialang.org/t/mathoptinterface-and-upcoming-breaking-changes-in-jump-0-19/4874). Since there is quite a big difference between JuMP 0.18 and JuMP 0.19, it makes sense to re-write `SDDP.jl` to support both Julia 1.0 and JuMP 0.19. This will also allow us to fix many outstanding issues (e.g., #10, #29, #68, #70, #79, #80, #84, #92, #107, #116; in fact, it will close almost all outstanding issues...).

That re-write is happening over at https://github.com/odow/Kokako.jl.

It still has a few notable things missing, including cut selection and parallelisation, but otherwise it already solves many of the issues mentioned above.

Kokako currently depends on the JuMP master branch and may change or break at any moment, so if you're risk averse, I recommend staying with SDDP.jl. However, if you want to join the future (and potentially run into loads of bugs), take Kokako out for a spin.

Since `SDDP.jl` is beginning to gain some brand recognition, it seems a shame to move over to Kokako permanently. Thus, once JuMP 0.19 is released and cut selection is implemented, I will merge Kokako into SDDP.jl as v2.0.

This will necessitate some updates to existing models. However, since it will occur in conjunction with the JuMP *and* Julia update, I feel it's better to lump everything in at once.

Please reply below if you have any queries/comments,
Oscar

### LINKED ISSUE #175: About dual multipliers of a constraint
Hi Oscar:

I have been trying to get the dual multipliers of several constraints in the first stage problem.

I have tried to store the the cuts (using the **cut_output_file** flag in the solving procedure), then create the first stage model using SDDP, load the cuts, then select the first stage problem, then write it to an **.lp** file using **JuMP.writeLP**. Then I have to load the saved problem (first stage + cuts) problem and solve it using CVX.

I have been reading 

> https://github.com/odow/SDDP.jl/pull/125

and

> https://github.com/odow/SDDP.jl/pull/125/commits/af0a4a62b6d866e40ca14521eec70d42460d4a58

But I can't figure it out a simpler way to get the dual multipliers. Is there any simpler way?

Kind Regards

### LINKED ISSUE #176: 2-dim tuple for noise
Dear Oscar, 

I'm trying to implement a process for 2 variables in the price interpolation scheme. 
Specifically I want to have a noise with a simultaneous but different movement for each variable. 
The DiscreteDistribution only allows a vector (N,1) of posibble values for the N possible realizations. But I need to specify a matrix of (N,2) size to include all posibble values. Is it somehow possible  to create a noise for DynamicPriceInterpolation in this case? 

Thanks!

### LINKED ISSUE #182: Installing SDDP
When I try to install by typing `] add https://github.com/odow/SDDP.jl#merge_kokako`
I get
```julia
  Updating git-repo `https://github.com/odow/SDDP.jl`
 Resolving package versions...
ERROR: Unsatisfiable requirements detected for package MathOptFormat [54bc58d1]:
 MathOptFormat [54bc58d1] log:
 ├─MathOptFormat [54bc58d1] has no known versions!
 └─restricted to versions * by SDDP [f4570300] — no versions left
   └─SDDP [f4570300] log:
     ├─possible versions are: 0.0.0 or uninstalled
     └─SDDP [f4570300] is fixed to version 0.0.0
```
Running `Pkg.installed()` gives
```
julia> Pkg.installed()
Dict{String,Union{Nothing, VersionNumber}} with 6 entries:
  "GLPK"             => v"0.9.1"
  "Compat"           => v"1.5.1"
  "Clp"              => v"0.6.1"
  "MathOptFormat"    => v"0.0.0"
  "MathOptInterface" => v"0.8.4"
  "JuMP"             => v"0.19.0"
```

Edit(odow): formatting

### LINKED ISSUE #183: SDDP.simulate with vectors
I am trying to use the simple LinearGraphPolicy. The example in the Kokako tutorial works. However, if I use a multidimensional control variable instead of only single dimensional control variables, I get the error
```
ERROR: LoadError: MethodError: no method matching value(::Array{VariableRef,1})
```
when I run the command 
```
simulations = SDDP.simulate(model, 1, [:controlvar, ]).
```
Running simulations = SDDP.simulate(model, 1) works fine, but then of course I cannot access the solution. Am I misunderstanding something?

### LINKED ISSUE #184: Saving trained model
Is there a way to save a trained model?

odow:
Just a small diff.

This PR branches off the existing SDDP.jl into a `release-v0` branch, and grafts on the existing [Kokako.jl](https://github.com/odow/Kokako.jl) repository in its place.

Closes #6
Closes #10
Closes #28
Closes #29
Closes #36
Closes #62
Closes #70
Closes #79
Closes #80
Closes #84
Closes #88
Closes #92
Closes #96
Closes #107
Closes #115
Closes #118
Closes #124
Closes #149
Closes #150
Closes #156
Closes #158
Closes #162
Closes #166
Closes #172
Closes #175
Closes #176
Closes #182
Closes #183
Closes #184


odow:
Once CI finishes I'm going to hit merge and then deal with the resulting fallout of people running `Pkg.update()` on 0.6.

PR #209, opened 2019-05-12
Add MSPPy hydro-thermal example

odow:
cc @lingquant: Are you okay with this under MPL? Or would you prefer BSD? It's not clear from the [file in msppy](https://github.com/lingquant/msppy/blob/dc85a2e8fa5243b3d5096d59085d9caad3ff2ede/examples/hydro_thermal/julia/test.jl) since you haven't updated the header.


odow:
In light of https://github.com/lingquant/msppy/pull/1#issuecomment-494428116, I'm going to merge this. I'm happy to change the license at a later date.

PR #218, opened 2019-05-19
Fix absolute value representation and update bound

bfpc:
The lower bound now matches the one in the paper (the convex problem is
solved to optimality)


odow:
Oops. I wondered why things were different.

PR #219, opened 2019-05-20
Add a settable optimize_hook with access to the current scenario.

odow:


peraaslid:
Can I use this to add new restriction before/after each iteration? I'd like to add DC powerflow active losses as "cuts" iterative. Could you give an example of how to use the pre-/post-optimize hook?

odow:
The current example is here:
https://github.com/odow/SDDP.jl/blob/9b5ebc986fb8a1a621c556c4852a4e6618dc5d7f/test/algorithm.jl#L150-L176
For each node in the tree, you can add a hook. To loop through nodes, you can go
```Julia
for (index, node) in model.nodes
    # add hook
end
```

`pre_optimize` hook is a function that takes the same arguments as `solve_subproblem`. If it returns a value, and `post_optimize_hook` is set, then the returned value will be passed as a the sole argument to `post_optimize_hook`.


```Julia
function my_pre_optimize_hook(model, node, state, noise, scenario_path, require_duals)
    sp = node.subproblem
    active_loss_constraint = @constraint(sp, sp[:x] <= 1) 
    return active_loss_constraint
end

function my_post_optimize_hook(ret)
    # `ret` is the `active_loss_constraint` reference returned from `my_pre_optimize_hook`
    println("Active loss constraint = $(ret)")
    return
end

for (index, node) in model.nodes
    SDDP.pre_optimize_hook(my_pre_optimize_hook, node)
    SDDP.post_optimize_hook(my_post_optimize_hook, node)
end
```

Note that `post_optimize_hook` doesn't get called immediately after `JuMP.optimize!`. I need that for my own work, but I could be compelled to change it if you need.

PR #224, opened 2019-05-26
Rp/backward

roporte:
Added samplers.jl. Backward_pass_sampler parameter added to SDDP.train. By default CompleteSampler() used.


src/plugins/samplers.jl
```diff
@@ -0,0 +1,27 @@
...
+
+    n_samples = min(length(node.noise_terms),mcs.number_of_samples)
+    sampled_t = sample(terms,n_samples)
+    u=unique(sampled_t)
+    d=Dict([(i,count(x->x==i,sampled_t)) for i in u])
+
+    for i in 1:length(u)
+        push!(sampled,Noise(u[i],d[u[i]]/n_samples))
+    end
+    return sampled
```
odow: You might want to use the [sample_noise](https://github.com/odow/SDDP.jl/blob/33e7024d93d6e20bd96a7428e5968ee347a885de/src/plugins/sampling_schemes.jl#L43-L60) function, so that you can go something like:
```Julia
function sample_backward_noise_terms(sampler::MonteCarloSampler, node)
    prob = 1 / sampler.number_of_samples
    return [
        Noise(
            sample_noise(InSampleMonteCarlo(), node.noise_terms),
            probability
        ) for _ in 1:sampler.number_of_samples
    ]
end
```
Although this doesn't guarantee that all samples are unique.

roporte: I tried to reuse InSampleMontecarlo(), but didnt find a way to guararantee uniqueness. I'd like to guarantee uniqueness because when the set of all noises is small, probably some noises are not unique repeated and same subprolem is solved more than once.

odow: We cache the subproblem solves, so it doesn't matter it the set isn't unique :) 

See https://github.com/odow/SDDP.jl/blob/b8d7ece3252c380194e630755950e88ff5a0cd10/src/algorithm.jl#L533-L540

But, if you want uniqueness, then we need to be cleverer. The easiest way is probably to duplicate the `sample_noise` function and pass it a set of noise terms that have been sampled.

Maybe something like (I haven't tested this):
```Julia
function sample_unique_noise(noise_terms::Vector{<:Noise}, samples::Set)
    if length(noise_terms) == 0
        return nothing
    elseif length(noise_terms) == length(samples)
        return nothing. # All terms have been sampled, no unique terms left to sample.
    end
    cumulative_probability = sum(noise.probability for noise in noise_terms)
    if cumulative_probability > 1.0 + 1e-6
        error("Cumulative probability cannot be greater than 1.0.")
    end
    rnd = rand() * cumulative_probability
    for noise in noise_terms
        rnd -= noise.probability
        if noise.term in samples
            continue
        end
        if rnd <= 0.0
            return noise.term
        end
    end
    error("Internal SDDP error.")
end
```

roporte: I didnt noticed that subproblems were cached! In this case i think your first suggestion is elegant, and reuse InSampleMontecarlo. Anyway ill try both solutions and compare.

src/SDDP.jl
```diff
@@ -19,6 +19,9 @@ include("user_interface.jl")
 # Default definitions for SDDP related modular utilities.
 include("plugins/headers.jl")
 
+# Default definitions for SDDP related modular utilities.
+include("plugins/samplers.jl")
```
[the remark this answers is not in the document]
roporte: Done.

src/plugins/samplers.jl
```diff
@@ -0,0 +1,27 @@
+include("../user_interface.jl")
```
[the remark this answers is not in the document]
roporte: Done

src/plugins/samplers.jl
```diff
@@ -0,0 +1,27 @@
...
+using StatsBase
```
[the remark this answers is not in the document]
roporte: Done

src/plugins/samplers.jl
```diff
@@ -0,0 +1,27 @@
...
+
+abstract type AbstractBackwardPassSampler end
```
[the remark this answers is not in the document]
roporte: i added some doc

src/plugins/samplers.jl
```diff
@@ -0,0 +1,27 @@
...
+
+struct CompleteSampler <: AbstractBackwardPassSampler end
+
+struct MonteCarloSampler <: AbstractBackwardPassSampler
```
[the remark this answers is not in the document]
roporte: Added.

src/plugins/headers.jl
```diff
@@ -140,3 +140,24 @@ function convergence_test(graph::PolicyGraph, log::Vector{Log},
...
+"""
+    AbstractBackwardPassSampler
+
+The abstract type for backward samplers interface.
+
+You need to define the following methods:
+ - [`SDDP.sample_backward_noise_terms`](@ref)
+
+"""
+
```
odow: Remove this blank line. The docstrings should be touching the line of code.

src/SDDP.jl
```diff
@@ -19,6 +19,7 @@ include("user_interface.jl")
 # Default definitions for SDDP related modular utilities.
 include("plugins/headers.jl")
 
+
```
odow: Minor comment: please remove this new line

src/algorithm.jl
```diff
@@ -456,7 +457,7 @@ function backward_pass(
                 belief == 0.0 && continue
                 solve_all_children(
                     model, model[node_index], items, belief, belief_state,
-                    objective_state, outgoing_state)
+                    objective_state, outgoing_state,backward_pass_sampler)
```
odow: Small comment: please uses spaces after `,`, so 
```
outgoing_state, backward_pass_sampler
```
instead of 
```
outgoing_state,backward_pass_sampler
```

src/algorithm.jl
```diff
@@ -480,7 +481,7 @@ function backward_pass(
             end
             solve_all_children(
                 model, node, items, 1.0, belief_state, objective_state,
-                outgoing_state)
+                outgoing_state,backward_pass_sampler)
```
odow: Ditto space :)

src/algorithm.jl
```diff
@@ -526,42 +527,48 @@ end
...
-                    update_objective_state(
-                        child_node.objective_state, objective_state, noise.term)
-                end
-                TimerOutputs.@timeit SDDP_TIMER "solve_subproblem" begin
-                    subproblem_results = solve_subproblem(
-                        model, child_node, outgoing_state, noise.term,
-                        require_duals = true)
+        if (child.probability > 0)
+            child_node = model[child.term]
+            sampled_noises = sample_backward_noise_terms(backward_pass_sampler,child_node)
```
odow: Ditto space :)

src/plugins/samplers.jl
```diff
@@ -0,0 +1,39 @@
+import StatsBase
+
+
+"""
+     CompleteSampler()
+
+ Backward sampler that returns all noises of the cosrresponding node.
```
odow: Spelling mistake: `corresponding` not `cosrresponding`. Also, please remove the blank line below.

src/plugins/samplers.jl
```diff
@@ -0,0 +1,39 @@
...
+struct CompleteSampler <: AbstractBackwardPassSampler end
+sample_backward_noise_terms(::CompleteSampler, node) = node.noise_terms
+
+
+"""
+     MonteCarloSampler(number_of_samples ::Integer)
+
+ Backward sampler that returns `number_of_samples` noises sampled
+ with replacement from noises on the corresponding node.
+
```
odow: Ditto blank line.

src/plugins/samplers.jl
odow: Okay. We could always add two sampling methods: one that preserves uniqueness, and the other that doesn't.

aferragu: A word of caution here: ensuring uniqueness is **not** the way Monte Carlo is supposed to work. If the noise distribution is far from uniform then by sampling without replacement you are introducing a bias in the expectation estimation.

For instance: if the noise probabilities are [0.97 0.01 0.01] then when you sample 3 of them it is highly likely that the first noise is sampled every time, or two out of three times. You have to take that into accoun in the averaging (i.e. consider the first noise with prob. 1 or 2/3 depending on the count).

If you sample 3 occurrences and ensure uniqueness then you are just averaging all the subproblems without regard for the original probabilities.

Granted, if you have only three noises you won't sample, but this is just to illustrate how it may fail when the noise distribution is not uniform.

If the noises are uniform, then both samplers are unbiased, but that's just a particular case.

I understand that you don't want to solve the subproblem twice. The solutions there are using the cached results (as @odow suggested), or creating the noise vector with *unique* supports but with probabilities number_of_times_sampled/number_of_samples. I favor this second approach.

src/algorithm.jl
roporte: done

src/plugins/samplers.jl
roporte: done

src/plugins/samplers.jl
roporte: done

src/algorithm.jl
roporte: done

src/algorithm.jl
roporte: done

src/SDDP.jl
roporte: done

src/plugins/headers.jl
roporte: done

src/algorithm.jl
```diff
@@ -587,24 +594,26 @@ function calculate_bound(model::PolicyGraph{T},
...
-                update_objective_state(node.objective_state,
-                    node.objective_state.initial_value, noise.term)
-            end
-            # Update belief state, etc.
-            if node.belief_state !== nothing
-                belief = node.belief_state::BeliefState{T}
-                partition_index = belief.partition_index
-                belief.updater(
-                    belief.belief, current_belief, partition_index, noise.term)
+        if (child.probability > 0)
```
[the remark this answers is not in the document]
roporte: done

src/algorithm.jl
```diff
@@ -526,42 +527,48 @@ end
...
-                end
-                if objective_state !== nothing
-                    update_objective_state(
-                        child_node.objective_state, objective_state, noise.term)
-                end
-                TimerOutputs.@timeit SDDP_TIMER "solve_subproblem" begin
-                    subproblem_results = solve_subproblem(
-                        model, child_node, outgoing_state, noise.term,
-                        require_duals = true)
+        if (child.probability > 0)
```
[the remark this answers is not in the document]
roporte: done

odow:
Do you know how to rebase things using git? I've been making changes, so we need to rebase onto the latest master. You can also resolve the conflicts using the web interface. If you don't, I can do it for you.

p.s. I'll be offline from tomorrow morning until Wednesday.

odow:
Hi @roporte, I took the liberty of tidying up your implementation. For now, if just delegated to the `InSampleMonteCarlo` as the sampler. If you want to implement something more clever, you can make another pull request. Thanks for your help!

PR #233, opened 2019-06-13
Remove Manifest.toml now MOF is registered

odow:
Ref https://github.com/odow/MathOptFormat.jl/issues/69


odow:
Tests failing because of #234.

PR #237, opened 2019-06-19
WIP: SDDiP

### LINKED ISSUE #195: SDDiP
Currently, an implementation of SDDiP lives in https://github.com/lkapelevich/SDDiP.jl.

I've had some discussions with @lkapelevich about bringing it into the fold with first-class support.

lkapelevich:
- [x] handle integer state variables (bug somewhere atm)
- [x] handle continuous state variables
- [x] remove all the `ext[:issddip]=true` and find a nice way to distinguish for a node/subproblem that we want Lagrangian duals (half done)
- [x] improve the code for calculating duals, ~~better initial values~~, get rid of extra vectors being created
- [x] remove manifest
- [x] move over remaining examples
- [x] pass the SDDP model's solver through to Kelley's method
- [x] remove the speed stuff from the binary expansion code, unnecessary
- [x] ~~raise an issue: state variables are assumed to be nonnegative here~~ just mentioned in docstrings
- [x] allow user to set epsilon in continuous state variables
closes #195 and https://github.com/lkapelevich/SDDiP.jl/issues/21


examples/air_conditioning.jl
```diff
@@ -20,11 +20,19 @@
...
     model = SDDP.LinearPolicyGraph(
             stages = 3, lower_bound = 0.0,
             optimizer = with_optimizer(GLPK.Optimizer)) do sp, stage
-        @variable(sp, stored_production >= 0, SDDP.State, initial_value = 0)
+
+        if sddip
+            sp.ext[:issddip] = true
+            state_ub = 100
+        else
+            state_ub = NaN
```
odow: You can do this!?! What does GLPK say?

src/user_interface.jl
```diff
@@ -749,27 +761,54 @@ function JuMP.build_variable(
...
             subproblem, JuMP.ScalarVariable(state_info.out), name * "_out")
     )
-    node = get_node(subproblem)
-    sym_name = Symbol(name)
-    @assert !haskey(node.states, sym_name)  # JuMP prevents duplicate names.
-    node.states[sym_name] = state
-    graph = get_policy_graph(subproblem)
-    graph.initial_root_state[sym_name] = state_info.initial_value
+
+    if !haskey(subproblem.ext, :issddip) || state_info.out.binary
```
odow: So the easiest way to do this is probably to create a `SDDP.SDDiPState()` and a `SDDiPStateInfo` and overload a new version of `JuMP.add_variable` which adds things appropriately, and makes sure to set `subproblem.ext[:isddip]`.

odow:
> handle integer state variables (bug somewhere atm)
> handle continuous state variables

```Julia
@variable(sp, x, SDDP.SDDiP.BinaryState(), initial_value = 1.0)
@variable(sp, 0 <= x <= 10, SDDP.SDDiP.ContinuousState(), initial_value = 1.0)
@variable(sp, 0 <= x <= 10, SDDP.SDDiP.IntegerState(), initial_value = 1.0)
```

> remove all the ext[:issddip]=true and find a nice way to distinguish for a node/subproblem that we want Lagrangian duals
  
See previous comment. Make the new JuMP variable types and set `:issddip` in `add_variable`. 

Bonus points for including a way to check for `SDDP.State` variables that have been added and throw an error if `:issddip` is set.

> improve the code for calculating duals, better initial values, get rid of extra vectors being created
 
👍 

> remove manifest

Why is it needed at the moment?

> move over remaining examples

More the merrier.

examples/booking_management.jl
```diff
@@ -13,43 +13,48 @@
...
+
+        if sddip
+            sp.ext[:issddip] = true
+        end
+
         @variable(sp,
-            0 <= vacancy[room=1:NUMBER_OF_ROOMS, day=1:NUMBER_OF_DAYS] <= 1,
-            SDDP.State, initial_value = 1)
+            0 <= vacancy[room=1:num_rooms, day=1:num_days] <= 1,
+            SDDP.State, Bin, initial_value = 1)
```
odow: Does this actually work?  I didn't realize...

examples/air_conditioning.jl
lkapelevich: nope haha I didn't get to running this earlier

lkapelevich:
> Why is it needed at the moment?

It's not, I pushed it without realizing

examples/booking_management.jl
lkapelevich: I took it for granted haha... so the state-outs will be binary, we don't care about the state-ins not actually being binary vars since they are always fixed

lkapelevich:
So by the `SDDP.SDDiP.ContinuousState()`, I take it you imagining this being a plugin/submodule? I think everything here can easily be moved, except the most invasive part here: https://github.com/odow/SDDP.jl/blob/67378f22d7aa69afb0cf54e5930a86852bc21d28/src/algorithm.jl#L167 
which relies on SDDP knowing about SDDiP, and I couldn't think of how to get this working without changing code `algorithm.jl`, or else parameterizing `Node`. So would you parameterize `Node` by the type of states it has (SDDP vs SDDiP)? 

There is also minor gotcha here https://github.com/odow/SDDP.jl/blob/67378f22d7aa69afb0cf54e5930a86852bc21d28/src/algorithm.jl#L868. 
but basically we just need to tell a Node whether it will be involved in SDDiP.

odow:
It doesn't have to be a submodule. You could also go `SDDP.SDDiPState()`. But having almost everything self-contained in a single file would be nice.

I don't really want to parameterize `Node`. Instead, one argument for including SDDiP in the package is so that we can have tighter interactions like you describe instead of splitting everything into lots of different packages. That said, you might be able to make an over-loadable version of `get_dual_variables` somehow.

lkapelevich:
After a think, I don't think there is a reason to introduce `SDDiPState` or similar unless you really like it as the interface. A user will either use SDDiP, in which case all variables need to be binary-fied, or they will not. So it's easier to just have a single switch for any state variables added, a keyword argument `SDDiP = true` in PolicyGraph will do everything necessary I think.

odow:
Is it possible to add the keyword to train? Then the default remains as-is, and we can solve the LP relaxation. You could even pass in a type, add it to `Options` and then use it to dispatch on any methods you need to overload.
```Julia
abstract type AbstractSMIPSolver end
struct ContinuousRelaxation <: AbstractSMIPSolver end
struct SDDiP <: AbstractSMIPSolver end
struct MyCoolNewSolver <: AbstractSMIPSolver end

SDDP.train(model, integrality_method = SDDP.SDDiP())
```

lkapelevich:
Oops I pushed before reading the comment

lkapelevich:
What exactly is the problem you see if we try to solve multiple times? If the keyword is added to `train` then the approach would be to not modify `add_variable` but instead in `train`, loop over all nodes and add new binary variables, then proceed with sddp. We would need to flag nodes once the extra variables have been added so that if we train multiple times the variables don't get added again. Were you thinking of something else?

odow:
Basically I would worry that someone would go
```Julia
SDDP.train(model, integrality_method = SDDP.SDDiP(), iteration_limit = 10)
SDDP.train(model, integrality_method = SDDP.ContinuousRelaxation(), iteration_limit = 10)
```
Would anything problematic happen the second time?

lkapelevich:
Yes this is actually worse than what I was thinking, because the first train would add a bunch of new state variables that we don't want for the second train.

examples/generation_expansion.jl
```diff
@@ -0,0 +1,51 @@
...
+capacities = ones(num_units)
+demand_vals = [
+    5 5 5 5 5 5 5 5
+    4 3 1 3 0 9 8 17
+    0 9 4 2 19 19 13 7
+    25 11 4 14 4 6 15 12
+    6 7 5 3 8 4 17 13
+    ]
+penalty = 5e5 # for not meething demand
+rho = 0.99 # discounting rate
```
odow: Please move all data into the function and add the standard license header

examples/vehicle_location.jl
```diff
@@ -41,7 +41,8 @@ function vehicle_location_model()
 
     model = SDDP.LinearPolicyGraph(
             stages = 10, lower_bound = 0.0,
-            optimizer = with_optimizer(GLPK.Optimizer)) do sp, t
+            optimizer = with_optimizer(GLPK.Optimizer),
+            mip_solver = SDDP.SDDiP() do sp, t
```
odow: Looks like you're missing a closing bracket here.

src/MIP_utils.jl
```diff
@@ -0,0 +1,101 @@
+#  Copyright 2017-19, Oscar Dowson.
```
odow: `, Lea Kapelevich.` or `, and contributors.`

src/MIP_utils.jl
```diff
@@ -0,0 +1,101 @@
...
+#  This Source Code Form is subject to the terms of the Mozilla Public
+#  License, v. 2.0. If a copy of the MPL was not distributed with this
+#  file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+
+abstract type AbstractMIPSolver end
+
+struct ContinuousRelaxation <: AbstractMIPSolver end
+
+mutable struct SDDiP <: AbstractMIPSolver
```
odow: Add docstrings.

src/user_interface.jl
```diff
@@ -749,27 +754,63 @@ function JuMP.build_variable(
             false, NaN,  # upper bound
             false, NaN,  # fixed value
             false, NaN,  # start value
-            false, false # binary and integer
+            false, false, # binary and integer, state in doesn't need integrality constraints
+            # info.binary, info.integer # binary and integer
```
odow: What's going on here? Can the comments be deleted and left as they were?

src/MIP_utils.jl
```diff
@@ -0,0 +1,101 @@
+#  Copyright 2017-19, Oscar Dowson.
+#  This Source Code Form is subject to the terms of the Mozilla Public
+#  License, v. 2.0. If a copy of the MPL was not distributed with this
+#  file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+
+abstract type AbstractMIPSolver end
```
odow: This should be added to `plugins/headers.jl`, along with the methods that need to be implemented by new subtypes (e.g., `get_dual_variables`).

Also, bikeshedding, this could be called `AbstractIntegralityHandler` (unless you have something better), and the file renamed to `plugins/integrality_handlers.jl`. It would also mean that the keyword argument to `PolicyGraph` could be `integrality_handler = SDDiP()`.

examples/stochastic_all_blacks.jl
```diff
@@ -0,0 +1,35 @@
+using SDDP, GLPK, Test
```
odow: Place in function, license header, etc.

examples/sddip.jl
```diff
@@ -0,0 +1,7 @@
+# TODO remove me
+
+include("air_conditioning.jl")
+include("all_blacks.jl")
+include("stochastic_all_blacks.jl")
+include("booking_management.jl")
+include("generation_expansion.jl")
```
odow: Any files in `examples` automatically get run in the tests.

src/algorithm.jl
```diff
@@ -271,6 +256,9 @@ function solve_subproblem(
     end
 
     JuMP.optimize!(node.subproblem)
+    state = get_outgoing_state(node)
+    stage_objective = stage_objective_value(node.stage_objective)
+    objective = JuMP.objective_value(node.subproblem)
```
odow: Is it important that these appear before the `get_dual_variables`?

src/algorithm.jl
```diff
@@ -901,7 +882,11 @@ function train(
     end
 
     # Handle integrality
-    binaries, integers = relax_integrality(model)
+    if model.nodes[1].integrality_handler == ContinuousRelaxation()
```
odow: It might be better to have 
```Julia
binaries, integers = relax_integrality(model, model.nodes[1].integrality_handler)
```
so that we can dispatch on new handlers.

src/binary_expansion.jl
```diff
@@ -0,0 +1,66 @@
+#  Copyright 2017-19, Oscar Dowson, Lea Kapelevich.
+#  This Source Code Form is subject to the terms of the Mozilla Public
+#  License, v. 2.0. If a copy of the MPL was not distributed with this
+#  file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+const _log2inv = inv(log(2))
+_bitsrequired(x::Int) = floor(Int, log(x) * _log2inv) + 1
+_bitsrequired(x::Float64, eps::Float64 = 0.1) = _bitsrequired(round(Int, x / eps))
```
odow: Where is the two-argument version used?

src/plugins/headers.jl
```diff
@@ -156,3 +156,30 @@ Returns a `Vector{Noise}` of noises sampled from `node.noise_terms` using
...
+    AbstractIntegralityHandler
+
+The abstract type for the integrality handlers interface.
+"""
+abstract type AbstractIntegralityHandler end
+
+"""
+    update_integrality_handler!(
+        integrality_handler::AbstractIntegralityHandler,
+        optimizer::JuMP.OptimizerFactory, num_states::Int)
```
odow: What is `num_states`?

src/plugins/headers.jl
```diff
@@ -156,3 +156,30 @@ Returns a `Vector{Noise}` of noises sampled from `node.noise_terms` using
...
+
+Helper function to set up `integrality_handler`, allocating any necessary
+storage. Fallback returns integrality_handler.
```
odow: Which fallback?

src/plugins/headers.jl
```diff
@@ -156,3 +156,30 @@ Returns a `Vector{Noise}` of noises sampled from `node.noise_terms` using
...
+"""
+function update_integrality_handler! end
+
+"""
+    get_dual_variables(node::Node, integrality_handler::AbstractIntegralityHandler)
+
+Returns the values of the dual variables associated with the fixed incoming
+state variables into `node`.
```
odow: What is the return type? "Returns a `Dict{Symbol, Float64}` where the keys are the names of the state variables and the values are the dual variables associated with the fishing constraint."

src/plugins/integrality_handlers.jl
```diff
@@ -0,0 +1,193 @@
+#  Copyright 2017-19, Oscar Dowson, Lea Kapelevich.
+#  This Source Code Form is subject to the terms of the Mozilla Public
+#  License, v. 2.0. If a copy of the MPL was not distributed with this
+#  file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import LinearAlgebra.dot
+
+# ========================= Fallbacks ======================================== #
+update_integrality_handler!(integrality_handler::AbstractIntegralityHandler, ::JuMP.OptimizerFactory, ::Int) = integrality_handler
```
odow: If this is the default, it can go as the stub method in `/headers.jl`.

src/plugins/integrality_handlers.jl
```diff
@@ -0,0 +1,193 @@
...
+end
+
+# =========================== SDDiP ========================================== #
+
+"""
+    SDDiP(; max_iter::Int = 100)
+
+The SDDiP integrality handler introduced by Zhou, J., Ahmed, S., Sun, X.A. in
+Nested Decomposition of Multistage Stochastic Integer Programs with Binary State
+Variables (2016).
```
odow: This has been published.

Zou, J., Ahmed, S. & Sun, X.A. Math. Program. (2019) 175: 461. https://doi.org/10.1007/s10107-018-1249-5

src/plugins/integrality_handlers.jl
```diff
@@ -0,0 +1,193 @@
...
+        ref = JuMP.FixRef(state.in)
+        dual_values[name] = dual_sign * JuMP.dual(ref)
+    end
+    return dual_values
+end
+
+# =========================== SDDiP ========================================== #
+
+"""
+    SDDiP(; max_iter::Int = 100)
```
odow: Prefer `iteration_limit::Int` to match other instances in the code.

src/plugins/integrality_handlers.jl
```diff
@@ -0,0 +1,193 @@
...
+    solver_obj = JuMP.objective_value(node.subproblem)
+    try
+        kelley_obj = _kelley(node, dual_vars, integrality_handler)::Float64
+        @assert isapprox(solver_obj, kelley_obj, atol = 1e-8, rtol = 1e-8)
+    catch e
+        write_subproblem_to_file(node, "subproblem", throw_error = false)
+        rethrow(e)
+    end
+    for (i, name) in enumerate(keys(node.states))
+        dual_values[name] = -dual_vars[i]
```
odow: Does this account for the optimization sense?

src/plugins/integrality_handlers.jl
```diff
@@ -0,0 +1,193 @@
...
+end
+
+function get_dual_variables(node::Node, integrality_handler::SDDiP)
+    dual_values = Dict{Symbol, Float64}()
+    # TODO implement smart choice for initial duals
+    dual_vars = zeros(length(node.states))
+    solver_obj = JuMP.objective_value(node.subproblem)
+    try
+        kelley_obj = _kelley(node, dual_vars, integrality_handler)::Float64
+        @assert isapprox(solver_obj, kelley_obj, atol = 1e-8, rtol = 1e-8)
```
odow: Maybe add a TODO to allow the setting of these tolerances?

src/plugins/integrality_handlers.jl
```diff
@@ -0,0 +1,193 @@
...
+
+function _solve_primal!(subgradients::Vector{Float64}, node::Node, dual_vars::Vector{Float64}, slacks)
+    model = node.subproblem
+    old_obj = JuMP.objective_function(model)
+    # Set the Lagrangian the objective in the primal model
+    fact = (JuMP.objective_sense(model) == JuMP.MOI.MIN_SENSE ? 1 : -1)
+    new_obj = old_obj + fact * dot(dual_vars, slacks)
+    JuMP.set_objective_function(model, new_obj)
+    JuMP.optimize!(model)
+    lagrangian_obj = JuMP.objective_value(model) # TODO moi/solver issue? gurobi/cplex can't get objective after modification
```
odow: Do you have a simpler example of this MOI bug?

src/plugins/integrality_handlers.jl
```diff
@@ -0,0 +1,193 @@
...
+            for (i, (name, state)) in enumerate(node.states)
+                JuMP.fix(state.in, integrality_handler.old_rhs[i], force = true)
+            end
+
+            return best_actual
+        end
+        # Next iterate
+        dual_vars .= value.(x)
+    end
+    error("Could not solve for Lagrangian duals.")
```
odow: More descriptive error? Iteration limit exceeded...

src/user_interface.jl
```diff
@@ -325,6 +325,8 @@ mutable struct Node{T}
     # An over-loadable hook for the JuMP.optimize! function.
     pre_optimize_hook::Union{Nothing, Function}
     post_optimize_hook::Union{Nothing, Function}
+    # Approach for handling discrete variables.
+    integrality_handler # TODO either leave untyped or define ::AbstractIntegralityHandler
```
odow: We have now defined an abstract type

src/user_interface.jl
```diff
@@ -765,11 +770,43 @@ function JuMP.add_variable(
             subproblem, JuMP.ScalarVariable(state_info.out), name * "_out")
     )
     node = get_node(subproblem)
-    sym_name = Symbol(name)
-    @assert !haskey(node.states, sym_name)  # JuMP prevents duplicate names.
-    node.states[sym_name] = state
-    graph = get_policy_graph(subproblem)
-    graph.initial_root_state[sym_name] = state_info.initial_value
+    if node.integrality_handler == ContinuousRelaxation() || state_info.out.binary
```
odow: It would be better to add a dispatchable function here. Something like
```Julia
function modify_state_variables(integrality_handler, node::Node, state_info, name)
```

test/binary_expansion.jl
```diff
@@ -0,0 +1,48 @@
+#  Copyright 2017-19, Oscar Dowson, Lea Kapelevich.
+#  This Source Code Form is subject to the terms of the Mozilla Public
+#  License, v. 2.0. If a copy of the MPL was not distributed with this
+#  file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+using SDDP: binexpand, bincontract
+using Test
+int_len = floor(Int, log(typemax(Int)) / log(2))
```
odow: Why is this not in the `testset`?

src/user_interface.jl
```diff
@@ -765,11 +770,43 @@ function JuMP.add_variable(
...
+        # Only in this case we treat `state` as a real state variable
+        sym_name = Symbol(name)
+        @assert !haskey(node.states, sym_name)  # JuMP prevents duplicate names.
+        node.states[sym_name] = state
+        graph = get_policy_graph(subproblem)
+        graph.initial_root_state[sym_name] = state_info.initial_value
+    else
+        if !isfinite(state_info.out.upper_bound)
+            error("When using SDDiP, state variables require an upper bound.")
+        end
```
odow: Don't you also require a non-negative lower bound?

test/runtests.jl
```diff
@@ -11,8 +11,10 @@ end
 
 const EXCLUDED_EXAMPLES = [
     "inventory_management.jl",
+    "generation_expansion.jl", # too tough for GLPK
```
odow: This should be tested with the `ContinuousRelaxation`. Better to disable SDDiP in the example and leave a note.

test/runtests.jl
```diff
@@ -11,8 +11,10 @@ end
...
     "msppy_hydro_thermal.jl",
-    "tiger_problem.jl"
+    "tiger_problem.jl",
+    "vehice_location.jl" # too tough
```
odow: Ditto.

examples/sddip.jl
odow: Still need to remove this.

src/algorithm.jl
lkapelevich: Can you suggest a way to check which integrality handler is being used, given `model`? I am assuming all nodes have the same integrality handler type

lkapelevich: Less ugly than `last(first(model.nodes)).integrality_handler`?

src/plugins/integrality_handlers.jl
lkapelevich: Yes if there are no bugs. The optimization sense should be taken into account in _kelley. The Lagrangian duals come out opposite sign to the LP duals. This convention could be changed, I can add a TODO.

src/user_interface.jl
lkapelevich: It's defined in headers.jl which is not included before user_interface

src/plugins/integrality_handlers.jl
lkapelevich: I haven't tried triggering this without JuMP yet- https://gist.github.com/lkapelevich/ef661edda4ab33699143cd49847e973d

src/user_interface.jl
lkapelevich: No, the reason that we are requiring an upper bound is to infer how many binary variables to create for an integer/continuous variables.

src/plugins/integrality_handlers.jl
odow: This isn't a bug. It's expected behavior. Modifying the objective invalidates the stored solution, and JuMP no longer caches the solution.

src/user_interface.jl
odow: Feel free to move things around if it works.

src/plugins/integrality_handlers.jl
lkapelevich: Makes sense :-)

src/algorithm.jl
lkapelevich: `stage_objective` should be cached, SDDiP will modify the objective while calculating duals

test/runtests.jl
lkapelevich: In this example even the continuous relaxation struggles

odow: Can we simplify it then? I don't really want examples that kill GLPK.

src/plugins/integrality_handlers.jl
lkapelevich: actually does this mean that the behavior is incorrect with GLPK, if GLPK returns the old objective?

odow: I think it's more undefined behavior. Some solvers may, but we shouldn't rely on it.

src/user_interface.jl
lkapelevich: I left this unaddressed, I don't see an easy way to do this

odow:
Looks like you forgot to commit `src/JuMP.jl`.

odow:
Great! The doc-test failures aren't your fault. The other PR's are also failing. I'm trying to fix it.

src/plugins/integrality_handlers.jl
```diff
@@ -0,0 +1,323 @@
...
+            JuMP.@constraint(subproblem, state.out == bincontract([binary_vars[i].out for i in 1:num_vars]))
+        else
+            initial_value = binexpand(float(state_info.initial_value), float(state_info.out.upper_bound), 0.1)
+            num_vars = length(initial_value)
+
+            binary_vars = JuMP.@variable(
+                subproblem, [i in 1:num_vars], base_name = "_bin_" * name,
+                SDDP.State, Bin, initial_value = initial_value[i])
+
+            # TODO allow for user-specified epsilon in place of default precision 0.1
```
odow: Pass a keyword into `SDDiP` with the precision?

src/plugins/integrality_handlers.jl
```diff
@@ -0,0 +1,323 @@
...
+end
+
+function get_dual_variables(node::Node, integrality_handler::SDDiP)
+    dual_values = Dict{Symbol, Float64}()
+    # TODO implement smart choice for initial duals
+    dual_vars = zeros(length(node.states))
+    solver_obj = JuMP.objective_value(node.subproblem)
+    try
+        kelley_obj = _kelley(node, dual_vars, integrality_handler)::Float64
+        # TODO allow these tolerances to be modified
```
odow: Also pass keywords into `SDDiP`?

src/plugins/integrality_handlers.jl
lkapelevich: This relates to the trickiness I mentioned about continuous variables- you suggested not worrying about continuous state variables for now. I couldn't see a simple way of passing that keyword argument appropriately to `add_variable`. If we store epsilon in `StateInfo`, then we are unnecessarily adding a field that's not used unless we are doing SDDiP. But we also don't want to create a new types such as `SDDiPState` and `SDDiPStateInfo` (and make a user type `SDDiPState` for every state variable) because whether or not SDDiP is used should be given through the integrality handler, and not via variables.

lkapelevich: Continuous variables with SDDiP could be special-cased (in another pr if anyone really wants this?) so the syntax would be `@variable(sp, x, SDDP.StateExpanded, initial_value = 0, epsilon = 0.01)`

src/plugins/integrality_handlers.jl
lkapelevich: OK should be easy

src/plugins/integrality_handlers.jl
lkapelevich: and SDDiP stores algorithmic things that relate to the entire model, whereas the epsilons are specific to each variable

odow: What about adding a `kwargs::Any` field to `StateInfo`, so people can go
```Julia
@variable(sp, x, SDDP.State, initial_value=0, epsilon=0.01)
```

lkapelevich: I will try that, sounds good

odow:
If you rebase onto the latest master, the documentation should be fixed.

lkapelevich:
@odow anything else?

lkapelevich:
I don't want to spend time writing docs right now, but happy to answer questions for docs.

PR #240, opened 2019-07-15
Updated hydro valley example and tests

conrad-mac:
Hey Oscar,

I had a go at updating the hydro valley example as an exercise in getting better acquainted with Julia/JuMP/SDDP. I'm still learning so it may not be terribly elegant. I followed your upgrading guide and managed to get there in the end. The tests all seem to pass.

Cheers,

Conrad


examples/sddp.jl_not_updated/HydroValleys/hydro_valley.jl
```diff
@@ -74,43 +73,45 @@ function hydrovalleymodel(;
...
-                stages          = 3,
-                objective_bound = flipobj * 1e6,
-                markov_transition = transition,
-                risk_measure    = riskmeasure,
-                cut_oracle      = cutoracle,
-                solver          = ClpSolver()
-                                        ) do sp, stage, markov_state
+    m = SDDP.MarkovianPolicyGraph(
+            sense                 = sense
+            , lower_bound         = lower
```
odow: I know this is a pedantic thing, but here and throughout, can you keep the commas on the same line please?

examples/sddp.jl_not_updated/HydroValleys/hydro_valley_tests.jl
```diff
@@ -4,76 +4,67 @@
...
+SDDP.train(
+    cutselection_model
+    , iteration_limit = 10
+    , print_level = 0
+    , cut_deletion_minimum = 2
 )
-
-@test length(SDDP.getsubproblem(cutselection_model, 2, 1).linconstr) < 10 + 11
-@test isapprox(getbound(cutselection_model), 855.0, atol=1e-3)
+# @test length(SDDP.getsubproblem(cutselection_model, 2, 1).linconstr) < 10 + 11
```
odow: See: http://www.juliaopt.org/JuMP.jl/v0.19.2/constraints/#Accessing-constraints-from-a-model-1

The way to do this is something like (I haven't tested):
```Julia
@test num_constraints(
        cutselection_model[(2, 1)], 
        JuMP.GenericAffExpr{Float64,JuMP.VariableRef}, 
        MOI.GreaterThan{Float64}
    ) < 10
```

examples/sddp.jl_not_updated/HydroValleys/hydro_valley.jl
```diff
@@ -74,43 +73,45 @@ function hydrovalleymodel(;
...
+            , upper_bound         = upper
+            , transition_matrices = transition
+            , optimizer           = with_optimizer(Clp.Optimizer)) do subproblem, node
```
odow: Please change the solver to `GLPK.Optimizer`. We don't load `Clp` on the CI machine.

examples/sddp.jl_not_updated/HydroValleys/hydro_valley_tests.jl
conrad-mac: `num_constraints` takes a JuMP `Model` object so I had to pass it the `subproblem` element of the node. 

The following test passes but only because `num_constraints` returns 0 (is this correct?).

```julia
@test num_constraints(
        cutselection_model[(2, 1)].subproblem, 
        JuMP.GenericAffExpr{Float64,JuMP.VariableRef}, 
        MOI.GreaterThan{Float64}
    ) < 10
```

Looking inside the model object, I see the following w/r/t constraints (although, I admit, I don't really understand what the difference between each is):

```julia
`GenericAffExpr{Float64,VariableRef}`-in-`MathOptInterface.EqualTo{Float64}`: 5 constraints
`GenericAffExpr{Float64,VariableRef}`-in-`MathOptInterface.LessThan{Float64}`: 4 constraints
`VariableRef`-in-`MathOptInterface.LessThan{Float64}`: 9 constraints
`VariableRef`-in-`MathOptInterface.GreaterThan{Float64}`: 15 constraints
```

examples/sddp.jl_not_updated/HydroValleys/hydro_valley.jl
conrad-mac: Sure thing. I don't know why this has become my code style now :)

examples/sddp.jl_not_updated/HydroValleys/hydro_valley.jl
conrad-mac: Done

examples/sddp.jl_not_updated/HydroValleys/hydro_valley.jl
conrad-mac: > In addition, take a look at the other examples. We wrap everything in a function so we're not polluting the global namespace with lots of different `model`s lying around.

Most of the example is already inside a function. Should I chuck those two `struct`s in the function or do you more mean to include all the tests at the bottom of the function (and call the function at the end. And obviously do away with the separate tests script)?

examples/sddp.jl_not_updated/HydroValleys/hydro_valley_tests.jl
odow: Oh, I forgot it was a maximization. You need to use `LessThan{Float64}`.

The new version of JuMP is unique amongst modeling packages; we see constraints not as `a'x {<=|=|>=} b`, but as `function-in-set`. So `a'x <= b` is `a'x in (-inf, b]`, or the function `GenericAffExpr` in the set `LessThan`. 

The `VariableRef`-in-`X` are variable bounds.

Here are some slides that give a good overview: https://www.juliaopt.org/meetings/bordeaux2018/lubin.pdf

examples/sddp.jl_not_updated/HydroValleys/hydro_valley.jl
odow: Yes, just add the code currently in `hydro_valley_tests.jl` to a new function `test_hydro_valley` and place it at the bottom of this file, then call the function at the end of the file.

So this file would become `/examples/hydro_valley.jl` and look like:
```Julia
struct Reservoir
  ...
end

function hydrovalleymodel()
 ...
end

function test_hydro_valley_model()
    @test ...
end

test_hydro_valley_model()
```

examples/sddp.jl_not_updated/HydroValleys/hydro_valley_tests.jl
conrad-mac: Cool, will make time to look at those slides. 

The test is still not passing. The number of constraints seems to change each time I train the model (and some of them are larger than 10).  I have been using `Random.seed!(11111)` - should I expect the output from train to be the same when using the same seed?

odow: Hmm. Can you push your latest changes? It's probably fine to just delta this test. It's not a very reliable way to achieve what I intended, and it is already tested in other places.

conrad-mac: Have pushed changes.

odow:
Thanks @conrad-mac!

PR #242, opened 2019-08-21
Add Parallel Simulate

andrewrosemberg:


src/algorithm.jl
```diff
@@ -3,6 +3,8 @@
 #  License, v. 2.0. If a copy of the MPL was not distributed with this
 #  file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+using Distributed
```
odow: Move this to `src/SDDP.jl` and change to an `import Distributed`

src/algorithm.jl
```diff
@@ -1129,7 +1131,8 @@ function simulate(model::PolicyGraph,
                       InSampleMonteCarlo(),
                   custom_recorders = Dict{Symbol, Function}(),
                   require_duals::Bool = true)
-    return map(i -> _simulate(
+    wp = CachingPool(workers())
+    return pmap(wp, (i) -> _simulate(
```
odow: I think you need a `let model = model` here somewhere: https://github.com/odow/SDDP.jl/pull/110/commits/0995948e548df39ee93d2dda00779c21d4404317.

We should also have 
```Julia
function simulate(
    parallel_workers = workers()
)
    if length(parallel_workers) < 2 
        return map(i -> _simulate(...))
    else
        wp = Distributed.CachingPool(parallel_workers)
        let model = model
            return pmap(wp, i -> _simulate(...))
        end
    end
end
```

odow: We also need a test with multiple workers. E.g., 

```Julia
results = simulate(model, 10, custom_recorders = Dict{Symbol, Function}(
    :myid => (args...) -> myid()
))
```
Then check that we get multiple different values in the `:myid` field.

test/algorithm.jl
```diff
@@ -269,3 +269,22 @@ end
...
+            ) do sp, t
+        @variable(sp, x[i=1:2] >= i, SDDP.State, initial_value = 2i)
+        @stageobjective(sp, x[1].out + x[2].out)
+    end
+    
+    np = size(Distributed.workers(),1)
+    results = SDDP.simulate(model, np, custom_recorders = Dict{Symbol, Function}(
+        :myid => (args...) -> Distributed.myid()
+    ))
+    @test size(unique([results[s][1][:myid] for s=1:np]),1) == np
```
odow: We need to remove the added procs at the end of this test.

test/algorithm.jl
```diff
@@ -269,3 +269,22 @@ end
         @test JuMP.upper_bound(node.subproblem[:i3]) == 2
     end
 end
+
+@testset "simulate parallel" begin
+    Distributed.addprocs()
+    @everywhere import Pkg; Pkg.activate(".")
```
odow: We should probably add `Distributed` as a dependency, and provide a helper function to add procs, using this work-around: https://discourse.julialang.org/t/packages-and-workers/14072/10?u=odow

People aren't going to like having to add this line.

src/algorithm.jl
```diff
@@ -1128,9 +1128,21 @@ function simulate(model::PolicyGraph,
                   sampling_scheme::AbstractSamplingScheme =
                       InSampleMonteCarlo(),
                   custom_recorders = Dict{Symbol, Function}(),
-                  require_duals::Bool = true)
-    return map(i -> _simulate(
+                  require_duals::Bool = true,
+                  parallel_workers = Distributed.workers())
+
+    if length(parallel_workers) < 2
```
odow: We should also add a check here to see if `model` contains subproblems in `direct_mode`. Because if they do, we should just use `map` instead of `pmap`. Also, add a `TODO` outlining why we need `direct_mode = false` (because the underlying C objects aren't copied, just the pointers).

PR #245, opened 2019-08-23
Fix custom_recorders example

Thuener:


odow:
Thanks! We should probably make these doctests to avoid these errors.

odow:
So the doc failure on Travis isn't your fault. But I'm not sure of the cause. Might be a documenter issue?

Thuener:
Maybe is the "open = true"?

odow:
I can't reproduce this locally. https://github.com/odow/SDDP.jl/pull/244 is incoming with a fix, but it might conflict with this, so I'll merge now and deal with the consequences in #244.

PR #265, opened 2019-10-14
Experiments with SDDiP

### LINKED ISSUE #264: SDDiP doesn't respect tolerances
The tolerances here should use the ones inside `SDDiP` instead of hard-coded `1e-8`:

https://github.com/odow/SDDP.jl/blob/951c5bf0ebc149c42465d37baee0189a84278c73/src/plugins/integrality_handlers.jl#L217

_Reported by Xian Yu, U. Mich._

odow:
Updates to integrality_handlers.jl.

Sorry for the big diff, I fixed some style points and had to remove unicode to get tests passing.

Closes #264


src/plugins/integrality_handlers.jl
```diff
@@ -170,51 +182,62 @@ function update_integrality_handler!(
...
 
 function get_dual_variables(node::Node, integrality_handler::SDDiP)
     dual_values = Dict{Symbol, Float64}()
-    # TODO implement smart choice for initial duals
+    # TODO: implement smart choice for initial duals
     dual_vars = zeros(length(node.states))
     solver_obj = JuMP.objective_value(node.subproblem)
     try
         kelley_obj = _kelley(node, dual_vars, integrality_handler)::Float64
-        @assert isapprox(solver_obj, kelley_obj, atol = 1e-8, rtol = 1e-8)
```
odow: @lkapelevich I've just removed this assertion. I can't think of a good way of checking this. The tolerances set by the user for dual feasibility imply this. We shouldn't be checking twice. If they choose a loose tolerance, that's on them. 

Unless there is there a way to make sure Kelly's always returns feasible duals?

odow: I've added back in but with much looser comparison.

odow:
@lkapelevich I've been playing around with SDDiP for the first (!) time today. 

Now that I understand it better: why Kelley's over a different method (e.g., level set)? MSPPY seems to have a good implementation: https://github.com/lingquant/msppy/blob/1ef95df4850deeee160b1e4456f10fc7e4d75080/msppy/sp.py#L1349

I tried playing around with regularizing the iterates etc., but in the high-dimensional problem of Xian's, the cutting planes really struggle.

odow:
Penalizes the L1 norm of the duals.

lkapelevich:
But I'm worrying about convergence in SDDP overall more than convergence inside the problem of finding the Lagrangian duals, which seems to be the issue in this problem.

lkapelevich:
Regarding favoring small absolute values- if e.g. you are minimizing and your cost to go is increasing in every direction you look, then the 0 vector is a feasible solution for the duals immediately. Or vice versa- you are maximizing and your cost to go is decreasing in every direction you look. Being in one of those situations is not super unlikely, that could be why favoring small absolute values is working well.

odow:
> that could be why favoring small absolute values is working well.

It also helps for numerical stability reasons. We want to avoid finding cuts that are very steep in one direction.

Can you comment on sub-gradient descent? It seems like it might help with Xian's problem where the value function is highly degenerate. Kelly's just pings back and forward, and we can also regularize the steps.

lkapelevich:
> the value function is highly degenerate

What do you mean by this?

I don't know what to expect in this problem, subgradient descent was usually worse than Kelley's and the level method when I played around with it. We should chat at informs.

odow:
> We should chat at informs.

Sunday

odow:
Other ideas to implement:
 - [ ] regularizing the duals by seeking the one that maximizes the height in the middle of the state-space (e.g., at `x=0.5`).
 - [ ] Don't binarize all states (thanks @cristianallara). We loose convergence, but anecdotally, it performs reasonably well.

odow:
@dukduque FYI

odow:
Closing since this is now stale. I have a new and revised plan for how to attack this. It means removing `IntegralityHandlers` and replacing them with modular cut handlers. This should lead to a more modular design for Benders cuts, Lagrangian cuts, and strengthened Benders cuts.

PR #282, opened 2019-12-16
Add support for approximating Markov lattice

### LINKED ISSUE #208: Scenario generation
@lingquant has some interesting things in msppy: [code](https://github.com/lingquant/msppy), [paper](http://www.optimization-online.org/DB_FILE/2019/05/7199.pdf).

In particular:

 - [ ] tools to help scenario generation
    We could integrate things like https://github.com/JuliaStochOpt/Scenarios.jl
    I've discussed this with Bernardo P. as well, basically given the analytic form of a process, 
    automatically discretize and form a policy graph. msppy shows that it's very helpful for users (#216)
 - [x] plot bound over time
    It'd be nice to throw up an interactive HTML dashboard with callbacks
 - [ ] the ability to simulate a continuous stochastic process
    We can actually do that now we have sampling oracles. Just needs to be implemented
 - [x] an extensive form solver
    Much like StochDynamicProgramming.jl
 - [x] parallelization
    We really need to re-implement this
 - [x] this stopping rule: https://github.com/odow/SDDP.jl/issues/178#issuecomment-491992499

odow:
This functionality is inspired by the ScenTrees.jl package
(https://github.com/kirui93/ScenTrees.jl).

Closes #208

Co-authored-by: Oscar Dowson <odow@users.noreply.github.com>
Co-authored-by: Kipngeno Kirui <kirui93@users.noreply.github.com>


odow:
@kirui93: for the short-term, I've implemented a slightly different version of your `LatticeApproximation` code, added you as a co-author to the Git commit so you will get credit on Github, and released the file under MIT instead of MPL2.

The main difference is that the probability matrices are normalized for each row, rather than each stage (i.e., the probabilities are conditioned on the stage and previous Markov state).

In the longer term, we can think about how to integrate `ScenTrees.jl` more broadly.

Somewhat novelly, we only ask the user for a single budget (the total number of nodes), and then we apportion these nodes to the stages based on the relative variance(ish) of each stage.

PR #297, opened 2020-01-23
Fix deterministic for affine expr with no terms

odow:
cc @haoxiangyang89 

This fixes the "K not defined." But we still throw on:
```julia
ERROR: Unable to formulate deterministic equivalent: `copy_and_replace_variables` is not implemented for functions like `-d[1]*G[1] + 0.5 d[1]*s[1,1] + t[0]_out - t[1]_out + 2.5 x[1,1]_out`.
```
because I didn't address quadratics yet.


odow:
Something weird is going on with coverage at the moment, because every line is hit, but it's marking the line below the hit line as missing... Merging even with the red x's.

PR #320, opened 2020-06-10
Add ForwardPass as a plug-in

### LINKED ISSUE #295: Forward pass plugins
There are two uses for this:

## Revisit old forward passes

In the serial world, we could run a few forward-backward iterations. Then stop. And perform a single backward pass, adding cut at every point in the state-space. For points that were sampled recently, this would have little effect (since they would have up-to-date information), but for points sampled a while ago, this would let us update their cuts without having to hit them again via sampling.

In the parallel case, it seems we should assign a processor to run backward-only passes using the set of states sampled to-date.

## Different forward models

The `SDDP.PolicyGraph` could be a convexification of an AC-OPF. This works fine for training, but if you dump the cuts into a full NLP AC-OPF and simulates, you often find that the solution visits points in the state-space that were not visited during training and are therefore sub-optimal. 

As a work-around, you could imagine training the convexification, dumping the cuts to the NLP, simulating 200 forward passes, the re-train the convexification using the 200 NLP forward passes.

## Action items

- [ ] Create a `AbstractForwardPass` plugin
- [ ] Refactor the current forward pass to `DefaultForwardPass <: AbstractForwardPass`
https://github.com/odow/SDDP.jl/blob/844cf46cc013dc223a1c8d3c03934f79bcc2db81/src/algorithm.jl#L347-L447

cc @joaquimg I think this is the easiest way forward. The difficulty is probably managing the transfer of the cuts. But you could even have two processors running and communicating via file, so the NLP and convexification are both running in parallel.

odow:
Closes #295 

@joaquimg does this seem like it's on the right track?


joaquimg:
Awesome thanks, I will test it

odow:
Get them to open an issue, and we can talk through the process for creating the NLP version. It should be pretty self-explanatory, though.

PR #333, opened 2020-07-16
fix bad spelling

lkapelevich:


odow:
Hilarious. I have a local branch on my computer where I was playing with some improvements to SDDiP. I think we have some good ideas.

lkapelevich:
I won't mind if you want to share :-)

odow:
I need to finish it and tidy it up, but basically we combine trust-region with Kelley's. Start with tight bounds on the duals (e.g., `-1e3 <= pi <= 1e3`). If Kelley's converges, a pi is at the boundary, and the state variables are violated, loosen the bound.

I think I'm also going to remove the special case integrality handlers and the binarization of the states. If users want to binarize their states; they should do it manually. We shouldn't encourage them.

I was doing this, then https://odow.github.io/StochOptFormat/ got in the way. Once we have that, we can run some proper benchmarks to test ideas.

PR #358, opened 2020-11-18
Minor typos

rtwalker:
Found a few typos while reading through docs and code


odow:
Thanks! Let me know when you finish and I'll merge.

rtwalker:
This is all I've got for now!

Though I noticed that there were a few dollar signs [here](https://odow.github.io/SDDP.jl/latest/examples/the_farmers_problem/) that seem like they have an extra escape character.

odow:
Ah. The extra dollar signs were needed at some point because Documenter was escaping weirdly. It looks like we can remove them, but it's not high priority.

PR #375, opened 2021-02-19
Update publication_plot.jl

### LINKED ISSUE #374: Minor warning annoyance when using Publication Plots
Hi Oscar:

I get a warning in Plots when using the Publication Plots recipe. Seems like the problem is in this line:
https://github.com/odow/SDDP.jl/blob/95e4a7fa9ed2e7ec91280abc7ceb793c23fc4b36/src/visualization/publication_plot.jl#L65

Changing alpha to seriesalpha should solve the issue.

Best,
Andres

odow:
Closes #374 

@aferragu is this what you meant? I didn't have time to check.

### DIFF
--- a/src/visualization/publication_plot.jl
+++ b/src/visualization/publication_plot.jl
@@ -62,7 +62,7 @@ RecipesBase.@recipe function f(
             ribbon := r
             y := μ
             fillalpha --> 0.2
-            alpha --> 0.0
+            seriesalpha --> 0.0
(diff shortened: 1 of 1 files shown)


aferragu:
Yep. Plots is throwing warnings if you do not use the proper parameter name (and not an alias) in the recipe. It was just annoying but when you have a jupyter notebook the warnings mess up the presentation....

PR #376, opened 2021-02-20
Update publication_plot.jl

odow:
@aferragu can you test this one?


aferragu (approved):
Looks great. Now tested with the same notebook. All warnings are gone. Thanks!!

PR #449, opened 2021-08-15
[breaking] Large refactoring of stochastic integer programs

### LINKED ISSUE #431: Use integer forward pass
I intended this. But at present we relax integrality before training, and re-instate after training. The default should be to enforce on the forward pass, and relax on the backward.

odow:
This is the start of a breaking PR for how we handle stochastic integer programs. I'm not in a rush to merge this so I plan on doing a big reorganization.

Plan of attack:

 - [x] Use integers on the forward pass
 - [x] Rename `ContinuousRelaxation` to `ConicDuality`
 - [x] Rename `SDDiP` to LagrangianDuality`
 - [x] Add `StrengthenedConicDuality`  
 - [x] Remove `integrality_handler` and let SDDP.jl decide when to try different dual cuts
 - [x] Remove automatic binarization and modify Kelley's as appropriate to accommodate non-binary states

Closes #431


src/plugins/duality_handlers.jl
```diff
@@ -0,0 +1,381 @@
+#  Copyright 2017-21, Oscar Dowson, Lea Kapelevich.
+#  This Source Code Form is subject to the terms of the Mozilla Public
+#  License, v. 2.0. If a copy of the MPL was not distributed with this
+#  file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+function _deprecate_integrality_handler()
+    return error(
+        """
+SDDP.jl v0.4.0 introduced a number of breaking changes in how we deal with
```
odow, line 9: cc @lkapelevich

PR #457, opened 2021-08-23
Fix bug in lagrangian duality

odow:
StrengthenedConicDuality can't use the higher bounds because the continuous dual solution is likely infeasible. If the bounds are too tight, then we'll get an incorrect dual objective value.

There must be something like adding a feasibility cut. But I don't know how that would work computationally. Using wide bounds seems okay.

cc @zfornier this fixes the StrengthenedConicDuality case, but not the LagrangianDuality. I obviously need to play with your example more.

### DIFF
--- a/src/algorithm.jl
+++ b/src/algorithm.jl
@@ -382,7 +382,9 @@ function solve_subproblem(
     end
     state = get_outgoing_state(node)
     stage_objective = stage_objective_value(node.stage_objective)
-    objective, dual_values = get_dual_solution(node, duality_handler)
+    TimerOutputs.@timeit SDDP_TIMER "get_dual_solution" begin
+        objective, dual_values = get_dual_solution(node, duality_handler)
+    end
     if node.post_optimize_hook !== nothing
         node.post_optimize_hook(pre_optimize_ret)
     end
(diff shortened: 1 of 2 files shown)


zfornier:
Since with the StrengthenedConicDuality option, SDDP finds a feasible
solution, respecting the binary constraint, I would like to understand
better how it works, how this parameter changes the way SDDP handles binary
variables. Would you have some articles in particular to recommend to me?

odow:
I should write a tutorial for the documentation.

I find how SDDiP is presented as a concept in the literature confusing. 

We have these subproblems: 
image
and we're trying to find a feasible dual solution and corresponding dual objective value for the slope and intercept of the cut respectively.

 * ContinuousConic relaxes the integrality and solves the conic dual https://odow.github.io/SDDP.jl/stable/apireference/#SDDP.ContinuousConicDuality
 * Strengthened conic duality uses the same dual solution vector, but strengthens it by computing a tighter dual objective value given that dual solution https://odow.github.io/SDDP.jl/stable/apireference/#SDDP.StrengthenedConicDuality
 * LagrangianDuality relaxes the fishing constraint (x-x=0) to compute the Lagrangian dual solution and corresponding objective value: https://odow.github.io/SDDP.jl/stable/apireference/#SDDP.LagrangianDuality

You should probably be able to find a feasible solution with the default `ContinuousConicDuality` option as well. I changed how we handle integer variables in the forward pass during training.

PR #468, opened 2021-09-05
Save states in write_cuts_to_file

### LINKED ISSUE #360: read_cuts_from_file does not load cuts into the BellmanFunction
Greetings,

While trying to visualize the value function of a model after reading cuts I had saved, I found out that was not possible. That is because `SDDP.read_cuts_from_file` doesn't add `SDDP.Cut` objects to the model's cut oracles. Training the model again whenever one wants to visualize the value function seems excessive.

I believe this could be solved in three different ways:
1. After calling `SDDP._add_cut`, create a `SDDP.Cut` object and `push!` it into the node's cut oracle. Functional, but inelegant because it repeats the process performed inside `SDDP._add_cut`;
2. Change `SDDP._add_cut` (or one of the functions called by it) so that it does the aforementioned process when reading cuts from file. It shouldn't break anything, as long as care is taken to avoid adding the same cut twice to the cut oracle;
3. Set `cut_selection = true` in the argument of `SDDP._add_cut` call inside `SDDP.read_cuts_from_file`. After testing proposed solution 1 and showing it to a professor, he made this suggestion. I tested it and it does add the cuts to the cut oracles. However, a comment in the function reads as follows:
    > `# Importantly, don't run cut selection when adding these cuts.`
    
    This leads me to believe you have a reason for not doing so, which I'm unaware of.

As mentioned above, I have tested 1 and 3 by creating functions outside of this module's scope and executing them. With my current setup, I can't create a branch of my local installation of SDDP and execute it. However, I could edit the local installation in a different machine and make a pull request, should you believe it to be necessary (although I personally see these changes as either minor or derived directly from your own code).

Thanks,
Vitor.

odow:
Closes #360 

Still needs some tests. 

@adow031, your [hack](https://github.com/odow/SDDP.jl/issues/360#issuecomment-912874789) isn't quite correct. It'll lead to cuts being added with the wrong intercept.

### DIFF
--- a/src/plugins/bellman_functions.jl
+++ b/src/plugins/bellman_functions.jl
@@ -579,23 +579,33 @@ function write_cuts_to_file(model::PolicyGraph{T}, filename::String) where {T}
         )
         oracle = node.bellman_function.global_theta.cut_oracle
         for (cut, state) in zip(oracle.cuts, oracle.states)
+            intercept = cut.intercept
+            for (key, π) in cut.coefficients
+                intercept += π * state.state[key]
+            end
             push!(
                 node_cuts["single_cuts"],
                 Dict(
-                    "intercept" => cut.intercept,
+                    "intercept" => intercept,
                     "coefficients" => copy(cut.coefficients),
                     "state" => copy(state.state),
                 ),
             )
         end
         for (i, theta) in enumerate(node.bellman_function.local_thetas)
-            for cut in theta.cut_oracle.cuts
+            oracle = theta.cut_oracle
+            for (cut, state) in zip(oracle.cuts, oracle.states)
+                intercept = cut.intercept
+                for (key, π) in cut.coefficients
+                    intercept += π * state.state[key]
+                end
                 push!(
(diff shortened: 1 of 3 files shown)


odow:
@adow031 how's this?

This is currently breaking if someone uses the `.json` file in a different workflow. I should probably make a `include_states::Bool = false` flag so that it's opt-in.

adow031:
The reason that I would have preferred that cuts just get written / read in the format that they are stored (i.e. theta is the intercept at x=0) is I have a model that has 36 states and needs 10s of thousands of cuts. 

This can take a while to load, so the modification of the intercept (back and forth) is just wasted time.

Also: there is a chance that this modification of theta could introduce numerical differences in the cuts, meaning that you get a different policy from saved cuts than from the internal model (but perhaps that can happen anyway).

I've written my work-around functions, so I may just end up using those, if I find there's is an appreciable difference in read/write times.

odow:
This still needs tests, and the option to not save the visited states.

PR #487, opened 2021-10-31
Use modified BFGS to solve Lagrangian

odow:
I need to run some proper benchmarks for this, but it seems much more stable than the previous code. I now understand the Lagrangian solve much better, and the trade-offs involved with computing the Lagrangian dual.

Given the extra complexity associated with it, it seems to me that we almost never want to use this duality handler. It's going to be orders of magnitude slower, particularly if we insist on computing optimal duals.

I've also done some refactoring, so we can now modify the `LocalImprovementSearch` module if we want to add more bells and whistles.

cc @lkapelevich you might be interested.


WalterMadelim:
I looked at your previous code.

Previously you attempted to solve the dual problem (which is about maximizing a concave function, with decision being `λ`) via a cutting plane method. 
```
L(λ) <= max t
         st t <= s * (L(λ_k) + h(x̄_k)' * (λ - λ_k)) for k=1,...
            t <= s * initial_bound
```
The idea of using cutting plane generation to solve it is very normal.
But the obstacle is that we even have no idea whether this concave maximization problem is bounded above! (So I find your `t <= s * initial_bound` confusing). Because this concave maximization problem is parametric (parameterized by the primal side candidate solution `x_trial`, which I see as the trickiest place).

And the unboundedness at the dual side typically is associated with infeasibility at the primal side. So whether one should generate feasibility cuts for the previous stages?

Therefore I think the biggest obstacle of using lagrangian duality is not in that we need to solve a sequence of MIP subproblems, but is in that the dual problem (i.e. maximizing the concave function) is too hard (or say, equivalently, to find a better dual solution i.e. the slope of the lagrangian cut, is too hard).

I don't quite understand if BFGS is suitable for this nondifferentiable concave maximization dual optimization case. Do you have any thoughts?

PR #511, opened 2022-02-25
Typo \cup-\cap

bfpc:
Hi Oscar.  I was reading the documentation on the new duality handlers (nice!) and found these typos.

### DIFF
--- a/src/plugins/duality_handlers.jl
+++ b/src/plugins/duality_handlers.jl
@@ -96,7 +96,7 @@ binary or integer restrictions as necessary.
 Given the problem
 ```
 min Cᵢ(x̄, u, w) + θᵢ
- st (x̄, x′, u) in Xᵢ(w) ∪ S
+ st (x̄, x′, u) in Xᵢ(w) ∩ S
     x̄ - x == 0          [λ]
 ```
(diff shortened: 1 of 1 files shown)


src/plugins/duality_handlers.jl
```diff
@@ -259,15 +259,15 @@ Obtain dual variables in the backward pass using strengthened conic duality.
 Given the problem
 ```
 min Cᵢ(x̄, u, w) + θᵢ
- st (x̄, x′, u) in Xᵢ(w) ∪ S
+ st (x̄, x′, u) in Xᵢ(w) ∩ S
     x̄ - x == 0          [λ]
 ```
 we first obtain an estiamte for `λ` using [`ContinuousConicDuality`](@ref).
```
odow, line 265:
```suggestion
we first obtain an estimate for `λ` using [`ContinuousConicDuality`](@ref).
```

odow (approved):
Yeah I think there's a lot of room for improvement on the question: what is the state-of-the-art for first-order non-smooth convex minimization? With a focus on few function evaluations, and being cheap to evaluate. 

Approximate methods are also okay, because we actually don't care about finding the global minimum.

PR #521, opened 2022-05-06
Fix broken link in air conditioning example

conema:
Substituted with a Wayback Machine link


odow:
Format check failure isn't your fault. I'll fix separately.

odow (approved):
A few unrelated failures that I obviously need to tidy up.

odow:
Thanks!

PR #531, opened 2022-11-22
Try to recover if we need dual solutions

bfpc:
Sometimes, the solver exits with a primal feasible point, but no dual solution.  If this happens during the backward pass, the algorithm exits, but it suffices (in the very few corner cases I could reproduce and test) to reset the solver just like if no primal feasible point was found: the `.mof.json` file produced by `SDDP.jl` indeed contains a perfectly feasible problem.

### DIFF
--- a/src/algorithm.jl
+++ b/src/algorithm.jl
@@ -354,6 +354,10 @@ function _uninitialize_solver(model::PolicyGraph; throw_error::Bool)
     return
 end
 
+# Internal function: decide whether we need a feasible dual point
+needs_duals(duality_handler::Nothing) = false
+needs_duals(duality_handler::AbstractDualityHandler) = true
+
 # Internal function: solve the subproblem associated with node given the
 # incoming state variables state and realization of the stagewise-independent
 # noise term noise.
@@ -390,6 +394,9 @@ function solve_subproblem(
         model.ext[:total_solves] = 1
     end
     if JuMP.primal_status(node.subproblem) != JuMP.MOI.FEASIBLE_POINT
+        attempt_numerical_recovery(node)
+    elseif needs_duals(duality_handler) &&
+           JuMP.dual_status(node.subproblem) != JuMP.MOI.FEASIBLE_POINT
         attempt_numerical_recovery(node)
     end
     state = get_outgoing_state(node)


src/algorithm.jl
```diff
@@ -389,7 +393,8 @@ function solve_subproblem(
     else
         model.ext[:total_solves] = 1
     end
-    if JuMP.primal_status(node.subproblem) != JuMP.MOI.FEASIBLE_POINT
+    if (JuMP.primal_status(node.subproblem) != JuMP.MOI.FEASIBLE_POINT) ||
+       (needs_duals(duality_handler) && JuMP.dual_status(node.subproblem) != JuMP.MOI.FEASIBLE_POINT)
```
odow, line 397: See the format failure. 

I think I'd prefer this to be a separate statement.

```Julia
if JuMP.primal_status(node.subproblem) != JuMP.MOI.FEASIBLE_POINT
    attempt_numerical_recovery(node)
elseif needs_duals(duality_handler) &&
       JuMP.dual_status(node.subproblem) != JuMP.MOI.FEASIBLE_POINT
    attempt_numerical_recovery(node)
```

It makes this a little clearer as to what is going on.

bfpc: It feels strange to me to have `attempt_numerical_recovery` twice, it could lead to changing one but not the other (and for me it seems less clear - my logic was "what are the conditions when we should try resolve".  The "format error" does not say explicitly what rule was broken (I suspect something along an 80-char limit given its solution).
I can see why you prefer your solution to the (awkward) parenthesis-hack, so I can change it if there's no way of bypassing the formatter for this one line (it stills seems more readable and clear to me to have one single if).

odow: The main reason to avoid complex `if` conditions is that it's hard to test and check that you've hit every branch. But I guess we're not testing this anyway because it's hard to trigger in CI.

If you want to keep the single statement, just copy the formatting from 

https://github.com/odow/SDDP.jl/actions/runs/3526117140/jobs/5913661433#step:4:200

bfpc: > The main reason to avoid complex `if` conditions is that it's hard to test and check that you've hit every branch. But I guess we're not testing this anyway because it's hard to trigger in CI.

Yeah, I can imagine :-)

BTW, this "resetting" trick was very nice to learn about.  On the other side, this brings (again) the question of having efficient (convex) NLP warm starts (I guess resetting only makes sense for LPs, or do you know of other examples?)

odow: > having efficient (convex) NLP warm starts

There are two parts to this: 
* if you use `@NL` macros, JuMP needs https://github.com/jump-dev/JuMP.jl/issues/1185
* if you're talking about warm-starting an interior point method, you can pass primal/dual starts to Ipopt, but there's still an algorithmic problem with the barrier.

odow: The resetting helps solvers which store things in-memory, like a factorization or a particular basis. Often, starting from scratch is more stable than attempting to warm-start from a previously solved solution.

PR #546, opened 2022-12-18
Revert "Try to recover if we need dual solutions (#531)"

odow:
This reverts commit 2eb30feaa18e551010398e55e4a8c59b1fbe8723.

@bfpc I'm reverting #531 because it doesn't account for cases with integer variables (so no dual solution will be present).

The place to fix it is probably here:

https://github.com/odow/SDDP.jl/blob/ed85827e85895ae7f3233d13a8bc4836db4fb56a/src/plugins/duality_handlers.jl#L112-L119

but that can be a separate PR.

### DIFF
--- a/src/algorithm.jl
+++ b/src/algorithm.jl
@@ -356,10 +356,6 @@ function _uninitialize_solver(model::PolicyGraph; throw_error::Bool)
     return
 end
 
-# Internal function: decide whether we need a feasible dual point
-needs_duals(duality_handler::Nothing) = false
-needs_duals(duality_handler::AbstractDualityHandler) = true
-
 # Internal function: solve the subproblem associated with node given the
 # incoming state variables state and realization of the stagewise-independent
 # noise term noise.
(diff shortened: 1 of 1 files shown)


bfpc:
> @bfpc I'm reverting #531 because it doesn't account for cases with integer variables (so no dual solution will be present).
> 
> The place to fix it is probably here:
> 
> https://github.com/odow/SDDP.jl/blob/ed85827e85895ae7f3233d13a8bc4836db4fb56a/src/plugins/duality_handlers.jl#L112-L119

I had thought about that, but I thought it might result in inconsistencies, since the primal solution found might not correspond to the dual solution if the latter gets recalculated.  Maybe the pair `(state, state_objective)` never gets used at the same time as `(objective, dual_values)`, but this is hard to formally impose.

Maybe one solution is to update the `needs_duals` to specialize to the ConicDuality (which does not re-solve the problem).  What do you think?

PR #547, opened 2022-12-18
Fix caching state bug in cyclic forward passes

### LINKED ISSUE #445: Starting states for infinite horizon
I've got an infinite horizon model with `terminate_on_cycle=true` and I've been looking at this section of the code.
```julia
        # ===== Begin: starting state for infinite horizon =====
        starting_states = options.starting_states[node_index]
        if length(starting_states) > 0
            # There is at least one other possible starting state. If our
            # incoming state is more than δ away from the other states, add it
            # as a possible starting state.
            if distance(starting_states, incoming_state_value) >
               options.cycle_discretization_delta
                push!(starting_states, incoming_state_value)
            end
            # TODO(odow):
            # - A better way of randomly sampling a starting state.
            # - Is is bad that we splice! here instead of just sampling? For
            #   convergence it is probably bad, since our list of possible
            #   starting states keeps changing, but from a computational
            #   perspective, we don't want to keep a list of discretized points
            #   in the state-space δ distance apart...
            incoming_state_value =
                splice!(starting_states, rand(1:length(starting_states)))
        end
        # ===== End: starting state for infinite horizon =====
```
I think that the best way to deal with steady-state is simply to start the next iteration using the ending state values from the current iteration, but this seems to be doing something a bit more complicated (and perhaps unnecessary, since the ending state values should reach a steady-state distribution over time, meaning you'll be sampling starting levels with the right distribution).

Also, since each iteration add at most one item to the `starting_states` vector and the `splice!` removes one value from the vector  it's not really clear it the list would ever be more than two items (the ending state, and the initial state), this would over-train using the initial state (unless I misunderstand the code).

Also, I would like to be able to set the `terminate_on_cycle=true` flag (which is needed for the ending states to be stored) when I'm running a historical simulation, is this possible?

Thanks,
Tony.

odow:
I think there was a very subtle bug in the implementation of the cyclic state-drop-off logic. We should be storing the outgoing state of the second-to-last node (which will be the incoming state of the node that forms a cycle), not the outgoing state of the node that forms a cycle.

Then JADE should use historical inflows which are 52weeks + 1week long and things should work by default.

We could potentially improve things by removing the `splice!`, so that we build up a distribution of starting states, not the trajectory over time.

Closes #445

### DIFF
--- a/src/plugins/forward_passes.jl
+++ b/src/plugins/forward_passes.jl
@@ -4,22 +4,36 @@
 #  file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 """
-    DefaultForwardPass()
+    DefaultForwardPass(; include_last_node::Bool = true)
 
 The default forward pass.
+
+If `include_last_node = false` and the sample terminated due to a cycle, then
+the last node (which forms the cycle) is omitted. This can be useful option to
+set when training, but it comes at the cost of not knowing which node formed the
+cycle (if there are multiple possibilities).
 """
-struct DefaultForwardPass <: AbstractForwardPass end
+struct DefaultForwardPass <: AbstractForwardPass
+    include_last_node::Bool
+    function DefaultForwardPass(; include_last_node::Bool = true)
+        return new(include_last_node)
+    end
+end
 
 function forward_pass(
     model::PolicyGraph{T},
     options::Options,
-    ::DefaultForwardPass,
+    pass::DefaultForwardPass,
 ) where {T}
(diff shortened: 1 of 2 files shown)


odow:
I think I need to check the logic of https://github.com/odow/SDDP.jl/issues/445 a bit more before merging.

odow:
Okay. I think I understand the various options.

This can be resolved in a non-breaking way by adding a `include_last_node::Bool = true` in the constructor of `DefaultForwardPass`, and a `terminates_on_cycle::Bool = false` to the constructor of `Historical`.

Issue #496, opened 2022-02-09
Documentation improvements


odow:
@bernardokp had these suggestions

<img width="564" alt="image" src="https://user-images.githubusercontent.com/8177701/153268333-857997e4-11b7-4401-9499-fd3b73309404.png">

<img width="833" alt="image" src="https://user-images.githubusercontent.com/8177701/153268498-f1408ec0-d6d0-466a-ac2a-de20442a0700.png">

<img width="360" alt="image" src="https://user-images.githubusercontent.com/8177701/153268275-7a8dcbc3-f2f6-4055-bae7-03760137631d.png">

x-ref #443

odow:
Moving #499 here

<img width="914" alt="image" src="https://user-images.githubusercontent.com/8177701/153515886-6cacfec7-09e9-4f3a-ac25-fd6e73d1fd4d.png">

odow:
From #516 



```
Doubts from the Vector auto-regressive model,
model = SDDP.LinearPolicyGraph(
    stages = 3,
    sense = :Min,
    lower_bound = 0.0,
    optimizer = GLPK.Optimizer,
) do sp, t
    @variable(sp, 0 <= x <= 200, SDDP.State, initial_value = 200)
    @variable(sp, g_t >= 0)
    @variable(sp, g_h >= 0)
    @variable(sp, s >= 0)
    @constraint(sp, g_h + g_t == 150)
    c = [50, 100, 150]
    @stageobjective(sp, c[t] * g_t)
    # =========================================================================
    # New stuff below Here
    # Add inflow as a state
    @variable(sp, inflow[1:2], SDDP.State, initial_value = 50.0)
    # Add the random variable as a control variable
    @variable(sp, ε[1:2])

    #why inflow[1:2] and ε[1:2] is written in this way and not inflow[2:1] and ε[2:1]?
    # Needed to transpose? Seems to me that the problem set up is 2x1 


This is a Julia thing. 2:1 means "the range from 2 to 1 in steps of +1", which is an empty set:

julia> x = 2:1
2:1

julia> collect(x)
Int64[]
 
You could write instead

julia> y = 2:-1:1
2:-1:1

julia> collect(y)
2-element Vector{Int64}:
 2
 1

which is "the range from 2 to 1 in steps of -1" and this is the set you expect. Also, inflow and ε are vectors. So they only have one dimension. That is, they aren't 2x1 matrices, but length-2 vectors:

julia> x = [1, 2]
[...]

PR #557, opened 2023-01-02
[docs] add two-stage newsvendor tutorial

odow:
Part of #496


odow:
Urgh. Nasty bound problems. It needed to be an upper bound on the second stage, not an upper bound on the full problem.

PR #579, opened 2023-04-01
Add numerical issue to log

odow:
@rafabench I was looking through the forks of SDDP.jl, and I saw these commits. Did you want to add them to upstream? I'll comment in-line on the changes.

### DIFF
--- a/src/algorithm.jl
+++ b/src/algorithm.jl
@@ -265,12 +265,12 @@ function attempt_numerical_recovery(model::PolicyGraph, node::Node)
             "creating the policy graph."
         )
     else
-        model.ext[:numerical_issue_counter] += 1
+        model.ext[:numerical_issue] = true
         MOI.Utilities.reset_optimizer(node.subproblem)
         optimize!(node.subproblem)
     end
     if JuMP.primal_status(node.subproblem) != JuMP.MOI.FEASIBLE_POINT
-        model.ext[:numerical_issue_counter] += 1
+        model.ext[:numerical_issue] = true
         @info "Writing cuts to the file `model.cuts.json`"
         write_cuts_to_file(model, "model.cuts.json")
         write_subproblem_to_file(
@@ -392,9 +392,6 @@ function solve_subproblem(
         model.ext[:total_solves] += 1
     else
         model.ext[:total_solves] = 1
-    end
-    if !haskey(model.ext, :numerical_issue_counter)
-        model.ext[:numerical_issue_counter] = 0
     end
     if JuMP.primal_status(node.subproblem) == JuMP.MOI.INTERRUPTED
         # If the solver was interrupted, the user probably hit CTRL+C but the
@@ -792,6 +789,7 @@ struct IterationResult{T}
 end
 
 function iteration(model::PolicyGraph{T}, options::Options) where {T}
+    model.ext[:numerical_issue] = false
     TimerOutputs.@timeit SDDP_TIMER "forward_pass" begin
         forward_trajectory = forward_pass(model, options, options.forward_pass)
         options.forward_pass_callback(forward_trajectory)
@@ -819,7 +817,7 @@ function iteration(model::PolicyGraph{T}, options::Options) where {T}
             Distributed.myid(),
             model.ext[:total_solves],
             duality_log_key(options.duality_handler),
-            model.ext[:numerical_issue_counter],
+            model.ext[:numerical_issue],
         ),
     )
     has_converged, status =
--- a/src/algorithm.jl
+++ b/src/algorithm.jl
@@ -968,7 +968,7 @@ function train(
     end
     # Reset the TimerOutput.
     TimerOutputs.reset_timer!(SDDP_TIMER)
-    log_file_handle = open(log_file, "w+")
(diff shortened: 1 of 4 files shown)


src/algorithm.jl
```diff
@@ -962,7 +968,7 @@ function train(
     end
     # Reset the TimerOutput.
     TimerOutputs.reset_timer!(SDDP_TIMER)
-    log_file_handle = open(log_file, "a")
+    log_file_handle = open(log_file, "w+")
```
odow, line 971: Hmm. Yeah this could be a reasonable change. It does annoy me sometimes that the log file grows without end. But equally, I often found it useful to be able to scroll back to look at past solves, particularly when developing a model.

src/plugins/parallel_schemes.jl
```diff
@@ -8,6 +8,7 @@ function log_iteration(options)
     if options.print_level > 0 &&
        mod(length(options.log), options.log_frequency) == 0
         print_helper(print_iteration, options.log_file_handle, options.log[end])
+        flush(options.log_file_handle)
```
odow, line 11: This seems reasonable, irrespective of the other changes.

src/print.jl
```diff
@@ -134,7 +134,7 @@ end
 function print_iteration_header(io)
     println(
         io,
-        " Iteration    Simulation       Bound         Time (s)    Proc. ID   # Solves",
+        " Iteration    Simulation       Bound         Time (s)    Proc. ID   # Solves     Num. Issues",
```
odow, line 137: Do you have models where there are many (recoverable) issues each iteration? What if we just added a `†` to indicate that there were numerical issues that iteration?

rafabench:
> @rafabench I was looking through the forks of SDDP.jl, and I saw these commits. Did you want to add them to upstream? I'll comment in-line on the changes.

Sure! I will comment and make the changes accordingly

src/algorithm.jl
rafabench: I just added this to make it faster for what I was testing, but I believe the append is the right choice. It's better to have a long file than to lose information of past runs.

src/plugins/parallel_schemes.jl
rafabench: This is very useful when the model breaks

src/print.jl
rafabench: I just added this column to make a graph of the accumulated numerical issues of the training proccess. The model that I'm working with has something like 10 recoverable issues in the whole run. I like the idea of `†`.

rafabench:
@odow Now the log looks like this:
```julia
 Iteration    Simulation       Bound         Time (s)    Proc. ID   # Solves     Num. Issues
...
     1075    5.647183e+05   5.753849e+05   1.022955e+03          1     503100           
     1076    5.821613e+05   5.753850e+05   1.024549e+03          1     503568           
     1077    5.864158e+05   5.753851e+05   1.026190e+03          1     504036          †
     1078    5.778750e+05   5.753852e+05   1.027832e+03          1     504504           
     1079    5.638553e+05   5.753853e+05   1.029456e+03          1     504972           
     1080    5.881738e+05   5.753854e+05   1.031150e+03          1     505440           
...
Terminating training
  Status         : time_limit
  Total time (s) : 3.602998e+03
  Total solves   : 1011816
  Best bound     :  5.755383e+05
  Simulation CI  :  5.751018e+05 ± 4.996055e+02
  Num. issues    : 3
------------------------------------------------------------------------------
```

src/print.jl
odow: I'm going to be a pain. I tried hard to keep the width of the log to less than 80 chars. Can we just append the dagger to the line, instead of giving it a whole new column?

So something like this:
```
     1077    5.864158e+05   5.753851e+05   1.026190e+03          1     504036†
```

src/print.jl
```diff
@@ -172,6 +174,9 @@ function print_footer(io, training_results::TrainingResults)
     μ, σ =
         confidence_interval(map(l -> l.simulation_value, training_results.log))
     println(io, "  Simulation CI  : ", print_value(μ), " ±", print_value(σ))
+    all_iterations_num_issues =
+        getfield.(training_results.log, :serious_numerical_issue)
+    println(io, "  Num. issues    : ", sum(all_iterations_num_issues))
```
odow, line 179:
```suggestion
    num_issues = sum(l -> l.serious_numerical_issue, training_results.log)
    println(io, "  Num. issues    : ", num_issues)
```

src/print.jl
rafabench: Could we add just one space to make it more visible?
```
     1077    5.864158e+05   5.753851e+05   1.026190e+03          1     504036 †
```
Or add the dagger on the beginning of the line (without messing the identation)

```
†    1077    5.864158e+05   5.753851e+05   1.026190e+03          1     504036 
```

odow: I don't mind. What do you think is best? It's really an extra for experts thing, so it can be subtle. I don't think we need to make it obvious.

odow:
Closing in favor of https://github.com/odow/SDDP.jl/pull/586

Issue #599, opened 2023-05-01
Multithreading parallelism


odow:
The asynchronous solver currently uses multiprocessing. This causes a lot of excess RAM usage because we need N copies of the full problem.

If we introduced a lock to every subproblem, then we could have the asynchronous threads operating in parallel over a single `PolicyGraph`. We'd just need to be careful to always obtain a lock whenever we modified data inside a node. And potentially avoid some of the plugins that have state, like `PSRSamplingScheme`.

I don't have plans to implement this, just noting so that I can come back to this in future.

odow:
One blocker is https://github.com/odow/SDDP.jl/blob/e631d22979163bc321908d97fdab53bcd7cbe26f/src/plugins/duality_handlers.jl#L64-L80. We'd need to change this so that we did it node-by-node, instead of all in one go.

odow:
I have something in #758 that seems to work well.

PR #611, opened 2023-05-13
Add AlternativeForwardPass

### LINKED ISSUE #419: [Research idea] ForwardPass plugins
In https://github.com/odow/SDDP.jl/issues/295, we introduced forwardness plugins. The motivation is for @andrewrosemberg's experiments with different network designs.

## Why do we want different models

Transmission models are approximations of reality. If we train a model using a poor approximation, then we will obtain a poor policy. 

The tricky thing is that we _can_ solve the proper AC-OPF, but this is non-convex, so we can't use it in SDDP.jl.

What we really want to do is to simulate forward passes with the AC-OPF to obtain forward trajectories, and then refine the value functions on the backward pass using some convex approximation (e.g., DC with line losses).

It isn't sufficient to use DC on the forward pass because this will visit the "wrong" points in the state-space, and so the true AC simulation will be suboptimal.

## Previous attempts

At present, it's easy to build-and-train a model using the same model on the forward and backward passes (e.g., NFA-NFA or DC-DC). However,  you can't mix models, or use the nonlinear AC-OPF.

The previous code hacked around this https://github.com/andrewrosemberg/SDDP.jl/tree/forw_diff_back, but it's now well out-of-date.

## Proposed solution

The ideal solution to this is for SDDP.jl to have some notion of a separate models on the forward and backward pass. However, this is a pretty niche request, and would lead to double the memory usage. I'm not going to do this.

Instead, we can leverage the forward pass plugins as follows.

- Build a DC model and an AC model. 
- Train the DC model for N iterations.
- Write the cuts from DC to file and load them into the AC problem
- Simulate the AC problem N times
- Use those simulations as a plugin for the next training of the DC problem

This logic should be able to be encapsulated within a forward pass problem so that it appears pretty seamless.

## Code

Here's a quick sketch of the possible implementation:
```Julia
mutable struct RosembergForwardPass{T} <: SDDP.AbstractForwardPass
    forward_model::SDDP.PolicyGraph{T}
    batch_size::Int
    batches::Vector{Any}
    counter::Int

    function RosembergForwardPass(;
        forward_model::SDDP.PolicyGraph{T},
        batch_size::Int,
    ) where {T}
        return new{T}(forward_model, batch_size, Any[], 0)
    end
end

function SDDP.forward_pass(
    model::PolicyGraph,
    options::Options,
    fp::RosembergForwardPass,
)
    fp.counter += 1
    if fp.counter > length(fp.batches)
        empty!(fp.batches)
        _fetch_new_simulations(model, fp)
        fp.counter = 1
    end
    return batches[fp.counter]
end

function _fetch_new_simulations(
    model::SDDP.PolicyGraph,
    options::Options,
    fp::RosembergForwardPass,
)
    # Update the cuts
    # We probably need some extra stuff here. You only need to load cuts not already
    # added, etc. Alternatively, we could just rebuild the forward_model every time?
    SDDP.write_cuts_to_file(model, "cuts.csv")
    SDDP.read_cuts_from_file(fp.forward_model, "cuts.csv")
    # Get new forward passes
    fp.batches = [
        SDDP.forward_pass(fp.forward_model, options, SDDP.DefaultForwardPass())
        for _ in 1:fp.batch_size
    ]
    return
end
```

## Pros and cons

The benefit of this approach is that it is simple. 

The downsides are that:
- it requires two SDDP.PolicyGrpah models (although I don't see a way of avoiding this)
- Batching the passes may slow convergence, although given the nature of the experiment, that's not a high priority. You could set the batch size to 1, but that would just result in lots of file IO moving the cuts across.

odow:
Closes #419 

Needs

- [x] Docs
- [x] Tests
- [x] An AC-OPF/DC-OPF example

But yeah. We had all the infrastructure for this. It was trivial to implement.


src/alternative_forward.jl
```diff
@@ -0,0 +1,50 @@
...
+
+function master_loop(
+    scheme::AlternativeParallelScheme{T},
+    model::PolicyGraph{T},
+    options::Options,
+) where {T}
+    _initialize_solver(model; throw_error = false)
+    while true
+        result = iteration(model, options)
+        slave_update(scheme.model, result)
```
odow, line 34: This was the only real difference. I guess we could have a post-iteration callback that runs on the master thread.

src/alternative_forward.jl
```diff
@@ -0,0 +1,50 @@
...
+struct AlternativeForwardPass{T} <: AbstractForwardPass
+    model::PolicyGraph{T}
+end
+
+function forward_pass(
+    ::PolicyGraph{T},
+    options::Options,
+    pass::AlternativeForwardPass{T},
+) where {T}
+    return forward_pass(pass.model, options, DefaultForwardPass())
```
odow, line 15: The power of abstractions :smile:

andrewrosemberg: Nice! Would it make sense to make this a batch?  Not sure if the convergence would be easy, but it could be interesting for a few cases.

src/alternative_forward.jl
andrewrosemberg: Would it allow us to usethe `Asynchronous` parallel schemes together with the different forward? again not sure if this would make sense,  but could be an interesting experiment.

src/alternative_forward.jl
```diff
@@ -0,0 +1,50 @@
...
+        log_iteration(options)
+        if result.has_converged
+            return result.status
+        end
+    end
+    return
+end
+
+function train_with_forward_model(nonconvex, convex; kwargs...)
+    return SDDP.train(
```
andrewrosemberg, line 44:
```suggestion
    @assert !isempty(setdiff(keys(nonconvex.initial_root_state), keys(convex.initial_root_state)))
    @assert !isempty(setdiff(keys(convex.initial_root_state), keys(nonconvex.initial_root_state)))
    return SDDP.train(
```

andrewrosemberg: Does this make sense?

src/alternative_forward.jl
```diff
@@ -0,0 +1,50 @@
+#  Copyright (c) 2017-23, Oscar Dowson and SDDP.jl contributors, Lea Kapelevich.
+#  This Source Code Form is subject to the terms of the Mozilla Public
+#  License, v. 2.0. If a copy of the MPL was not distributed with this
+#  file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
```
andrewrosemberg, line 5:
```suggestion
"""
    AlternativeForwardPass(;
        forward_model::SDDP.PolicyGraph{T},
    )

A forward pass for holding an alternative foward model. This is useful for
calculating cuts and simulating the policy under different models.
For example, simulating the forward passes with the AC-OPF to obtain forward trajectories,
and then refine the value functions on the backward pass using some convex approximation 
(e.g., DC with line losses) as in the following paper:
 - Rosemberg, Andrew Rosemberg, Alexandre Street, Joaquim Dias Garcia, Davi M. Valladão, Thuener Silva, and Oscar Dowson.
"Assessing the cost of network simplifications in long-term hydrothermal dispatch planning models." 
IEEE Transactions on Sustainable Energy 13, no. 1 (2021): 196-206.
"""
```

andrewrosemberg:
Creating a AC-OPF/DC-OPF example. Will post here in a few minutes. Taking advantage of the code used in PowerModelsAnnex.jl.

andrewrosemberg:
[case3.zip](https://github.com/odow/SDDP.jl/files/11470468/case3.zip)

The ACDC test case. I used PowerModels.jl just to create the reference dictionary to make things a bit easier, but I can try to do it manually.

andrewrosemberg:
I added a few comments and the example based on personal interest, but if it confuses things, please disregard them.

src/alternative_forward.jl
odow: Changed. Potentially. I'm planning on replacing the asynchronous algorithm with https://github.com/odow/SDDP.jl/issues/599 which should work.

src/alternative_forward.jl
odow: Added ability to specify the forward pass within the forward pass.

odow:
Once I merge #612 and rebase, this might take the record for the most complicated feature implemented in the fewest lines of code. It's really only two one-line functions...

docs/src/examples/pglib_opf.jl
```diff
@@ -0,0 +1,127 @@
...
+        sense = :Min,
+        lower_bound = 0.0,
+        optimizer = Ipopt.Optimizer,
+    ) do sp, t
+        power_model = PowerModels.instantiate_model(
+            data,
+            model_type,
+            PowerModels.build_opf;
+            jump_model = sp,
+        )
```
odow, line 39: @andrewrosemberg I added `PowerModels` as a doc and test dependency instead. No need to complicate things for this example.

docs/src/examples/pglib_opf.jl
```diff
@@ -0,0 +1,127 @@
...
+
+# To summarizing, training with the convex model constructs cuts at points that
+# may never be visited by the non-convex model, and training with the non-convex
+# model may construct arbitraily poor cuts because a key assumption of SDDP is
+# convexity.
+
+# As a compromise, we can train a policy using a combination of the convex and
+# non-convex models; we'll use the non-convex model to generate trial points on
+# the forward pass, and we'll use the convex model to build cuts on the backward
+# pass.
```
odow, line 106: It's likely that this writing is super confusing to anyone who hasn't ready Andrew's paper... At some point we'll migrate this tutorial from `examples` to a proper tutorial, but it'll do for now.

src/alternative_forward.jl
```diff
@@ -0,0 +1,64 @@
...
+A post-iteration callback that should be used whenever [`SDDP.AlternativeForwardPass`](@ref)
+is used.
+"""
+struct AlternativePostIterationCallback{T}
+    model::PolicyGraph{T}
+end
+
+function (callback::AlternativePostIterationCallback)(result::IterationResult)
+    slave_update(callback.model, result)
```
odow, line 62: I'm super pleased with how trivial this turned out to be.

odow:
Provided two models have the same graph structure and state variables, we now have a lot of freedom:

 * Simulating with the true distribution, training with the discretized SAA (sampling schemes, https://github.com/odow/SDDP.jl/pull/594)
 * Simulating with the discrete model, training with the continuous relaxation (happens by default)
 * Training with a non-convex model, training with a convex approximation (this PR)

docs/src/examples/pglib_opf.jl
```diff
@@ -0,0 +1,127 @@
...
+# SDDP.jl may find a sub-optimal policy. Therefore, it may over-estimate the
+# true cost of operation.
+
+non_convex = build_model(PowerModels.ACPPowerModel)
+SDDP.train(non_convex; iteration_limit = 10)
+result = SDDP.simulate(non_convex, 1)
+
+# ## Combining convex and non-convex models
+
+# To summarizing, training with the convex model constructs cuts at points that
```
odow, line 98:
```suggestion
# To summarize, training with the convex model constructs cuts at points that
```

odow:
Preview: https://odow.github.io/SDDP.jl/previews/PR611/examples/pglib_opf/

odow:
@andrewrosemberg this tutorial is so cool: https://odow.github.io/SDDP.jl/previews/PR611/tutorial/pglib_opf/. 

It's like a mini-HydroPowerModels.jl.

These new features should make it easy for people to replicate and build upon your paper now.

PR #653, opened 2023-08-24
[MSPFormat] detect and exploit stagewise independent lattices

odow:


odow:
Seems okay, but I don't have enough examples to know for sure.

PR #662, opened 2023-09-07
Fix default stopping rule with limits

odow:
x-ref https://github.com/EPOC-NZ/JADE.jl/pull/44


src/algorithm.jl
```diff
@@ -1057,6 +1054,10 @@ function train(
     if time_limit !== nothing
         push!(stopping_rules, TimeLimit(time_limit))
     end
+    # If no stoppinng rule exists, add the default rule.
```
odow, line 1057:
```suggestion
    # If no stopping rule exists, add the default rule.
```

PR #677, opened 2023-09-24
[docs] update docs/src/tutorial/example_reservoir.jl

odow:


docs/src/tutorial/example_reservoir.jl
```diff
@@ -346,53 +335,95 @@ SDDP.plot(
...
-#     @stageobjective(sp, data[t, :cost] * u_thermal + terminal_cost_function)
-# else
-#     @stageobjective(sp, data[t, :cost] * u_thermal)
-# end
-# ```
+# This graph has two nodes, and a loop from node 2 back to node 1 with
+# probability 0.7.
 
-# ### Higher fidelity modeling
+# We can construct a cylic policy graph as follows:
```
odow, line 352:
```suggestion
# We can construct a cyclic policy graph as follows:
```

docs/src/tutorial/example_reservoir.jl
```diff
@@ -346,53 +335,95 @@ SDDP.plot(
...
 
-# Take this model and modify it following the suggestions below. The SDDP.jl
-# documentation has a range of similar examples and hints for how to achieve
-# them.
+# One major problem with our model is that the reservoir is empty at the end of
+# the time horizon. This is because our model does not consider the cost of
+# future years after the `T` weeks.
 
-# ### Terminal value functions
+# We can fix this using a cylic policy graph. One way to construct a graph is
```
odow, line 344:
```suggestion
# We can fix this using a cyclic policy graph. One way to construct a graph is
```

PR #695, opened 2023-10-09
Fix non-constant state dimension in MSPFormat reader

### LINKED ISSUE #693: MSPFormat: dimension of state variable can change between nodes
https://github.com/bonnkleiford/MSPLib-Library/issues/2#issuecomment-1743130867

odow:
Closes #693

### DIFF
--- a/src/MSPFormat.jl
+++ b/src/MSPFormat.jl
@@ -305,6 +305,17 @@ function read_from_file(
                 ω_objective[x] = objective
             end
         end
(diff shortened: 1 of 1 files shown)


odow:
This still can't read the files from https://github.com/bonnkleiford/MSPLib-Library/issues/2#issuecomment-1743130867 because there are other terms in the models that I don't know what to do with.

odow:
cc @bonnkleiford

PR #704, opened 2023-10-25
Add examples for MDP paper

odow:


odow:
Closing, but I will come back to this. See #720 instead.

PR #720, opened 2023-12-12
Add code for MDP paper

odow:


src/user_interface.jl
```diff
@@ -1411,6 +1411,15 @@ function construct_belief_update(
             end
             outgoing_belief[node_i] = PY_X * PX / PY
         end
+        if length(outgoing_belief) == 2
+            for (node_i, belief) in incoming_belief
+                if belief < 1e-6
```
odow, line 1416: I need to think about this. We encounter numerical issues as we approach the edge of the belief space

PR #742, opened 2024-05-13
Add sample_backward_noise_terms_with_state

### LINKED ISSUE #735: Error when using nonlinear function of state.in
I am trying to solve a problem that is dependent on a nonlinear function of both the incoming state (that sould be a parameter) and the uncertainty.

However, since both the uncertainty and the incoming state are treated as variables in JuMP I am getting the following error:

```
ERROR: MathOptInterface.UnsupportedConstraint{MathOptInterface.ScalarNonlinearFunction, MathOptInterface.GreaterThan{Float64}}: `MathOptInterface.ScalarNonlinearFunction`-in-`MathOptInterface.GreaterThan{Float64}` constraint is not supported by the model.
```
Here is a minimum working case based on one of SDDP.jl documentation examples, where we demonstrate the error on a trivial constraint. Is there a way around? 

```
using SDDP, Gurobi, Test

function fast_hydro_thermal()
    model = SDDP.LinearPolicyGraph(;
        stages = 2,
        upper_bound = 0.0,
        sense = :Max,
        optimizer = Gurobi.Optimizer,
    ) do sp, t
        @variable(sp, 0 <= x <= 8, SDDP.State, initial_value = 0.0)
        @variables(sp, begin
            y >= 0
            p >= 0
            ξ
        end)
        @constraints(sp, begin
            p + y >= 6
            x.out <= x.in - y + ξ
        end)
        RAINFALL = (t == 1 ? [6] : [2, 10])
        SDDP.parameterize(sp, RAINFALL) do ω
            return JuMP.fix(ξ, ω)
        end
        @stageobjective(sp, -5 * p)

        @constraint(sp, 1/x.in*exp(ξ) >= 0)

    end

   SDDP.train(model)
    return
end

fast_hydro_thermal()
```
For us this problem appears when we are implementing a specific stoachstic process for modeling the inflow inside the SDDP. Here are the equations for our case. 

```
    lower_bound = - (μ_stage) /  σ_stage - sum(ϕ[j]*inflow[i, end - j + 1] for j in 1:p)
    λ = (var_resid/lower_bound^2) + 1
    μ = 0.5 * log(var_resid/ (λ * (λ - 1)))
    σ = sqrt(log(λ))
    ϵ[i] = exp(ξ*σ+ μ) + lower_bound

    @constraint(subproblem, inflow_modeling[i = 1:n_hydro], inflow[i, 6].out == sum(ϕ[i, t][j]*inflow[i, end - j + 1].in for j in 1:p[i, t]) + ϵ[i])
```

Another way around would be to treat ϵ[i] as a parameter that is evaluated before solving the subproblem.

arthur-brigatto:
Closes #735.

### DIFF
--- a/src/algorithm.jl
+++ b/src/algorithm.jl
@@ -659,12 +659,11 @@ function solve_all_children(
             continue
         end
         child_node = model[child.term]
-        for noise in
-            sample_backward_noise_terms_with_state(
-                backward_sampling_scheme,
-                child_node,
-                outgoing_state,
-            )
+        for noise in sample_backward_noise_terms_with_state(
+            backward_sampling_scheme,
+            child_node,
+            outgoing_state,
+        )
             if length(scenario_path) == length_scenario_path
                 push!(scenario_path, (child.term, noise.term))
             else
--- a/src/algorithm.jl
+++ b/src/algorithm.jl
@@ -660,7 +660,11 @@ function solve_all_children(
         end
         child_node = model[child.term]
         for noise in
-            sample_backward_noise_terms(backward_sampling_scheme, child_node, outgoing_state)
+            sample_backward_noise_terms_with_state(
+                backward_sampling_scheme,
+                child_node,
+                outgoing_state,
+            )
             if length(scenario_path) == length_scenario_path
                 push!(scenario_path, (child.term, noise.term))
             else
--- a/src/plugins/backward_sampling_schemes.jl
+++ b/src/plugins/backward_sampling_schemes.jl
@@ -10,7 +10,7 @@ Backward sampler that returns all noises of the corresponding node.
 """
 struct CompleteSampler <: AbstractBackwardSamplingScheme end
 
-sample_backward_noise_terms(::CompleteSampler, node::Node) = node.noise_terms
+sample_backward_noise_terms(::CompleteSampler, node) = node.noise_terms
 
 """
      MonteCarloSampler(number_of_samples::Int)
--- a/src/plugins/backward_sampling_schemes.jl
+++ b/src/plugins/backward_sampling_schemes.jl
@@ -22,7 +22,7 @@ struct MonteCarloSampler <: AbstractBackwardSamplingScheme
     number_of_samples::Int
 end
 
-function sample_backward_noise_terms(sampler::MonteCarloSampler, node::Node, state::Dict{Symbol,Float64})
+function sample_backward_noise_terms(sampler::MonteCarloSampler, node::Node)
     prob = 1 / sampler.number_of_samples
     return [
         Noise(sample_noise(node.noise_terms), prob) for
--- a/src/plugins/backward_sampling_schemes.jl
+++ b/src/plugins/backward_sampling_schemes.jl
@@ -10,7 +10,7 @@ Backward sampler that returns all noises of the corresponding node.
 """
 struct CompleteSampler <: AbstractBackwardSamplingScheme end
 
-sample_backward_noise_terms_with_state(::CompleteSampler, node::Node, state::Dict{Symbol,Float64}) = node.noise_terms
+sample_backward_noise_terms(::CompleteSampler, node::Node) = node.noise_terms
 
 """
      MonteCarloSampler(number_of_samples::Int)
--- a/src/plugins/headers.jl
+++ b/src/plugins/headers.jl
@@ -126,7 +126,7 @@ function sample_backward_noise_terms end
         state::Dict{Symbol,Float64},
     )::Vector{Noise}
 
-Returns a `Vector{Noise}` of noises sampled conditionally on the sate `state` using
+Returns a `Vector{Noise}` of noises sampled conditionally on the `state` using
 `backward_sampling_scheme`.
 """
 function sample_backward_noise_terms_with_state(
--- a/src/plugins/headers.jl
+++ b/src/plugins/headers.jl
@@ -132,7 +132,7 @@ Returns a `Vector{Noise}` of noises sampled conditionally on the `state` using
 function sample_backward_noise_terms_with_state(
     sampler::AbstractBackwardSamplingScheme,
     node::Node,
-    state::Dict{Symbol,Float64},
+   ::Dict{Symbol,Float64},
 )
     return sample_backward_noise_terms(sampler, node)
 end
--- a/src/plugins/headers.jl
+++ b/src/plugins/headers.jl
@@ -112,6 +112,7 @@ abstract type AbstractBackwardSamplingScheme end
     sample_backward_noise_terms(
         backward_sampling_scheme::AbstractBackwardSamplingScheme,
         node::Node{T},
+        state::Dict{Symbol,Float64},
     )::Vector{Noise}
 
 Returns a `Vector{Noise}` of noises sampled from `node.noise_terms` using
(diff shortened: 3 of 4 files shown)


andrewrosemberg:
A simple test would be great to see if it works!

src/algorithm.jl
```diff
@@ -660,7 +660,7 @@ function solve_all_children(
         end
         child_node = model[child.term]
         for noise in
-            sample_backward_noise_terms(backward_sampling_scheme, child_node)
+            sample_backward_noise_terms(backward_sampling_scheme, child_node, outgoing_state)
```
odow, line 663:
```suggestion
            sample_backward_noise_terms_with_state(
                backward_sampling_scheme,
                child_node,
                outgoing_state,
            )
```

src/plugins/backward_sampling_schemes.jl
```diff
@@ -10,7 +10,7 @@ Backward sampler that returns all noises of the corresponding node.
 """
 struct CompleteSampler <: AbstractBackwardSamplingScheme end
 
-sample_backward_noise_terms(::CompleteSampler, node) = node.noise_terms
+sample_backward_noise_terms_with_state(::CompleteSampler, node::Node, state::Dict{Symbol,Float64}) = node.noise_terms
```
odow, line 13:
```suggestion
sample_backward_noise_terms(::CompleteSampler, node::Node) = node.noise_terms
```

src/plugins/backward_sampling_schemes.jl
```diff
@@ -22,7 +22,7 @@ struct MonteCarloSampler <: AbstractBackwardSamplingScheme
     number_of_samples::Int
 end
 
-function sample_backward_noise_terms(sampler::MonteCarloSampler, node::Node)
+function sample_backward_noise_terms(sampler::MonteCarloSampler, node::Node, state::Dict{Symbol,Float64})
```
odow, line 25:
```suggestion
function sample_backward_noise_terms(sampler::MonteCarloSampler, node::Node)
```

src/plugins/backward_sampling_schemes.jl
```diff
@@ -10,7 +10,7 @@ Backward sampler that returns all noises of the corresponding node.
 """
```
odow, line 10: I can't add a suggestion to the top of the file, but add the default implementation:
```julia
function sample_backward_noise_terms_with_state(
    sampler::AbstractBackwardSamplingScheme,
    node::Node,
    state::Dict{Symbol,Float64},
)
    return sample_backward_noise_terms(sampler, node)
end
```

odow: Actually, it should go here: https://github.com/odow/SDDP.jl/blob/8ed2e302c7212388ba3c4ec67c4737eaa44aeed5/src/plugins/headers.jl#L101-L120

odow:
You can add a new test as a function just after this one:
https://github.com/odow/SDDP.jl/blob/8ed2e302c7212388ba3c4ec67c4737eaa44aeed5/test/plugins/backward_sampling_schemes.jl#L66-L90

src/plugins/headers.jl
```diff
@@ -112,6 +112,7 @@ abstract type AbstractBackwardSamplingScheme end
     sample_backward_noise_terms(
         backward_sampling_scheme::AbstractBackwardSamplingScheme,
         node::Node{T},
+        state::Dict{Symbol,Float64},
```
odow, line 115: Okay, so:

I want two methods:

1. the existing `sample_backward_noise_terms`
2. a new `sample_backward_noise_terms_with_state`

as well as a fallback

```Julia
function sample_backward_noise_terms_with_state(
    sampler::AbstractBackwardSamplingScheme,
    node::Node,
    state::Dict{Symbol,Float64},
)
    return sample_backward_noise_terms(sampler, node)
end
```

The reason for this is to maintain backwards compatibility: we want all existing code to continue to work, but we want the ability for new code to use `sample_backward_noise_terms_with_state` instead.

odow, line 115:
```suggestion
```

arthur-brigatto: Alright, I had the wrong idea. I will correct it.

arthur-brigatto:
> You can add a new test as a function just after this one:
> 
> https://github.com/odow/SDDP.jl/blob/8ed2e302c7212388ba3c4ec67c4737eaa44aeed5/test/plugins/backward_sampling_schemes.jl#L66-L90

Working on it!

src/plugins/headers.jl
```diff
@@ -115,10 +115,28 @@ abstract type AbstractBackwardSamplingScheme end
...
 function sample_backward_noise_terms end
 
+"""
+    sample_backward_noise_terms_with_state(
+        sampler::AbstractBackwardSamplingScheme,
+        node::Node,
+        state::Dict{Symbol,Float64},
+    )::Vector{Noise}
+
+Returns a `Vector{Noise}` of noises sampled conditionally on the sate `state` using
```
odow, line 129:
```suggestion
Returns a `Vector{Noise}` of noises sampled conditionally on the `state` using
```

src/plugins/headers.jl
```diff
@@ -115,10 +115,28 @@ abstract type AbstractBackwardSamplingScheme end
...
+`backward_sampling_scheme`.
```
odow, line 130:
```suggestion
`sampler`.
```

arthur-brigatto:
@odow I have added a test function. The idea is: Firstly a sequence of 0.1 or 0.9 values for $\epsilon$ is sampled in the forward step. Then, the backward sampler sets $\epsilon = 0.9$ if 0.1 was sampled in the forward step or it sets $\epsilon = 0.1$ if 0.9 was sampled in the forward step.

Let me know if that works!

src/plugins/backward_sampling_schemes.jl
```diff
@@ -10,7 +10,7 @@ Backward sampler that returns all noises of the corresponding node.
...
 struct CompleteSampler <: AbstractBackwardSamplingScheme end
 
-sample_backward_noise_terms(::CompleteSampler, node) = node.noise_terms
+sample_backward_noise_terms(::CompleteSampler, node::Node) = node.noise_terms
```
odow, line 13:
```suggestion
sample_backward_noise_terms(::CompleteSampler, node) = node.noise_terms
```

src/plugins/headers.jl
```diff
@@ -115,10 +115,28 @@ abstract type AbstractBackwardSamplingScheme end
...
+        state::Dict{Symbol,Float64},
+    )::Vector{Noise}
+
+Returns a `Vector{Noise}` of noises sampled conditionally on the `state` using
+`sampler`.
+"""
+function sample_backward_noise_terms_with_state(
+    sampler::AbstractBackwardSamplingScheme,
+    node::Node,
+    state::Dict{Symbol,Float64},
```
odow, line 135:
```suggestion
   ::Dict{Symbol,Float64},
```

src/plugins/headers.jl
```diff
@@ -115,10 +115,28 @@ abstract type AbstractBackwardSamplingScheme end
...
+        state::Dict{Symbol,Float64},
+    )::Vector{Noise}
+
+Returns a `Vector{Noise}` of noises sampled conditionally on the `state` using
+`sampler`.
+"""
+function sample_backward_noise_terms_with_state(
+    sampler::AbstractBackwardSamplingScheme,
+    node::Node,
+   ::Dict{Symbol,Float64},
```
odow, line 135:
```suggestion
    ::Dict{Symbol,Float64},
```

PR #758, opened 2024-07-23
Add Threaded parallel scheme

### LINKED ISSUE #599: Multithreading parallelism

odow:
Closes #599

This is heading in the right direction:
```Julia
julia> @time simulations = SDDP.simulate(
           model,
           10_000;
           parallel_scheme = SDDP.Threaded(),
           custom_recorders = Dict{Symbol, Function}(
               :thread_id => sp -> Threads.threadid(),
           ),
       );
 14.560144 seconds (58.22 M allocations: 2.101 GiB, 5.02% gc time, 0.06% compilation time)

julia> @time simulations = SDDP.simulate(
           model,
           10_000;
           parallel_scheme = SDDP.Serial(),
           custom_recorders = Dict{Symbol, Function}(
               :thread_id => sp -> Threads.threadid(),
           ),
       );
 47.935609 seconds (58.31 M allocations: 2.106 GiB, 2.24% gc time, 0.10% compilation time)
```

Avoiding the various race conditions is a bit fiddly though.

## TODOs

 - [x] integer stuff?

### DIFF
--- a/test/plugins/threaded.jl
+++ b/test/plugins/threaded.jl
@@ -51,7 +51,7 @@ function test_threaded()
         custom_recorders = recorder,
     )
     thread_ids_seen =
-        Set{Int}(data[:thread_id] for sim in simulations, data in sim)
+        Set{Int}(data[:thread_id] for sim in simulations for data in sim)
     min_threads = Threads.nthreads() > 1 ? 1 : 2
     @test min_threads <= length(thread_ids_seen) <= Threads.nthreads()
     return
--- a/test/plugins/threaded.jl
+++ b/test/plugins/threaded.jl
@@ -41,7 +41,7 @@ function test_threaded()
     SDDP.train(model; iteration_limit = 100, parallel_scheme = SDDP.Threaded())
     thread_ids_seen =
         Set{Int}(log.pid for log in model.most_recent_training_results.log)
-    min_threads = Threads.nthreads() > 1 ? 1 : 2
+    min_threads = Threads.nthreads() == 1 ? 1 : 2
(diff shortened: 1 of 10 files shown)


odow:
So the issue is that `prepare_backward_pass` assumes that we will be doing _only_ forward passes or _only_ backward passes. We need to change this to apply per-node.

test/plugins/threaded.jl
```diff
@@ -0,0 +1,60 @@
...
+    @test min_threads <= length(thread_ids_seen) <= Threads.nthreads()
+    recorder = Dict{Symbol,Function}(:thread_id => sp -> Threads.threadid())
+    simulations = SDDP.simulate(
+        model,
+        100;
+        parallel_scheme = SDDP.Threaded(),
+        custom_recorders = recorder,
+    )
+    thread_ids_seen =
+        Set{Int}(data[:thread_id] for sim in simulations, data in sim)
```
odow, line 54:
```suggestion
        Set{Int}(data[:thread_id] for sim in simulations for data in sim)
```

odow:
There's still testing to be improved, but I think this is ready for basic usage.

PR #761, opened 2024-07-24
DNM: Fix various Threading() bugs

odow:


src/algorithm.jl
```diff
@@ -4,13 +4,18 @@
...
 
 macro _timeit_threadsafe(timer, label, block)
-    return esc(quote
-        if Threads.threadid() == 1
+    code = quote
+        first_thread = 1
+        if VERSION >= v"1.7"
+            first_thread = first(Threads.threadpooltids(Threads.threadpool()))
+        end
+        if Threads.threadid() == first_thread
```
dannyopts, line 12: Is this relevant here https://julialang.org/blog/2023/07/PSA-dont-use-threadid/ 

I believe that Threads.threadid() could change throughout the course of a threads execution and so multiple threads or no threads may have the id first_thread at this point.

src/algorithm.jl
```diff
@@ -302,11 +307,13 @@ function attempt_numerical_recovery(model::PolicyGraph, node::Node)
     end
     if !_has_primal_solution(node)
         model.ext[:numerical_issue] = true
-        @info "Writing cuts to the file `model.cuts.json`"
-        write_cuts_to_file(model, "model.cuts.json")
+        suffix = Threads.threadid() == 1 ? "" : "_pid=$(Threads.threadid())"
```
dannyopts, line 310: As above, I dont think this is safe, you could end up with multiple threads writing to the same file in either an overlapping or overwritting fashion. Maybe just a uuid instead of threadid would be safer?

src/algorithm.jl
```diff
@@ -869,20 +888,20 @@ function iteration(model::PolicyGraph{T}, options::Options) where {T}
...
+        has_converged, status =
+            convergence_test(model, options.log, options.stopping_rules)
+        return IterationResult(
+            max(Threads.threadid(), Distributed.myid()),
+            bound,
+            forward_trajectory.cumulative_value,
+            has_converged,
+            status,
+            cuts,
+            model.ext[:numerical_issue],
```
dannyopts, line 900: I think you could be writing to model.ext[:numerical_issue] from other threads at this point, also I think you could have overwritten the state from another thread between it being set to true within forward or backward pass and the time that you return it here

src/algorithm.jl
```diff
@@ -302,11 +307,13 @@ function attempt_numerical_recovery(model::PolicyGraph, node::Node)
     end
     if !_has_primal_solution(node)
         model.ext[:numerical_issue] = true
```
dannyopts, line 309: Do you need a lock here before writing to the model?

src/algorithm.jl
```diff
@@ -1399,7 +1433,7 @@ function simulate(
     custom_recorders = Dict{Symbol,Function}(),
     duality_handler::Union{Nothing,AbstractDualityHandler} = nothing,
     skip_undefined_variables::Bool = false,
-    parallel_scheme::AbstractParallelScheme = Serial(),
+    parallel_scheme::AbstractParallelScheme = Threaded(),
```
dannyopts, line 1436: docs still say "Defaults to [`Serial`]"

dannyopts:
Hey, I hope you dont mind me putting some comments on. Hopefully they are useful, if not then applogies :)

src/plugins/parallel_schemes.jl
```diff
@@ -362,14 +362,26 @@ function master_loop(
...
+                        lock(convergence_lock) do
+                            keep_iterating = false
+                            status = result.status
+                            return
+                        end
+                    end
+                end
+            finally
                 lock(convergence_lock) do
                     keep_iterating = false
```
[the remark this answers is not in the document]
odow: Yeah I'm aware of this. I had it as-is because it doesn't really matter if we read the "wrong" value of `keep_iterating`.

src/algorithm.jl
odow: The "DNM" in the title means "Do Not Merge" :smile: I don't plan to make this the default anytime soon, so I'll be reverting this line before merging.

src/algorithm.jl
odow: I don't particularly care about this one, since it is only for profiling with `TimerOutputs`. Perhaps I should just turn this off if `nthreads() > 1`.

src/algorithm.jl
odow: IIUC, tasks can be moved between threads, but if two tasks are on the same thread, they can't be running concurrently? But I guess the risk is that the task might be swapped between this line and the next?

Regardless, a UUID is a better idea.

src/algorithm.jl
odow: We can fix `total_solves` easily. 

The `:numerical_issue` is a bit more complicated, but to be honest, it doesn't really matter. It's just for reporting a very subtle thing in the log.

src/algorithm.jl
```diff
@@ -1308,7 +1342,7 @@ end
         custom_recorders = Dict{Symbol, Function}(),
         duality_handler::Union{Nothing,AbstractDualityHandler} = nothing,
         skip_undefined_variables::Bool = false,
-        parallel_scheme::AbstractParallelScheme = Serial(),
+        parallel_scheme::AbstractParallelScheme = Threaded(),
```
odow, line 1345:
```suggestion
        parallel_scheme::AbstractParallelScheme = Serial(),
```

odow:
Merging for now as an improvement. I need to fix some other things, but I'll do them in separate PRs.

Issue #777, opened 2024-09-02
Getting out of local optimal SDDiP


Thuener:
Hi Oscar, 

now with SDDiP, we actually have a heuristic to solve the model. We don't proved convergence anymore. I can see that on several occasions, the model is stuck on a local optimal. I have been using BanditDuality to help the model to get out of those local optimal. Using different cuts helps get out of the local optimal. However, StrengthenedConicDuality can make much slower progress.

Any ideas on how to get out of the local optimal?
Is there any way to use different duality_handlers depending on the bound stalling?

PS: I also get an issue with BanditDuality when trying to start back the train, it gets stuck with just one type of duality. Example:

```
     1883S -3.526655e+04  2.575658e+03  5.779006e+04    750405   1
      1885S -6.197976e+04  2.575544e+03  5.828998e+04    751967   1
      1887S -3.414856e+04  2.568755e+03  5.844276e+04    752881   1
      1889S -3.993974e+04  2.568373e+03  5.876231e+04    753719   1
      1890  -5.145811e+04  2.568373e+03  5.880455e+04    754762   1
      1891S -1.481099e+04  2.568067e+03  5.883862e+04    754869   1
      1893S -2.585710e+04  2.568097e+03  5.899596e+04    755583   1
      1895S -2.168459e+04  2.567502e+03  5.911727e+04    755921   1
      1897S -2.168718e+04  2.562634e+03  5.917999e+04    756171   1
      1899S -2.104896e+04  2.559852e+03  5.930590e+04    756765   1
      1900  -6.025239e+04  2.559852e+03  5.935892e+04    758124   1
      1901S -2.302431e+04  2.559217e+03  5.946921e+04    758411   1
      1903S -1.432304e+04  2.558576e+03  5.951825e+04    758661   1
      1905S -3.676079e+04  2.557187e+03  5.980848e+04    759567   1
      1909S -4.680980e+04  2.555810e+03  6.017820e+04    760987   1
-------------------------------------------------------------------
status         : time_limit
total time (s) : 6.017820e+04
total solves   : 760987
best bound     :  2.555810e+03
simulation ci  : -1.533718e+05 Â± 1.864513e+04
numeric issues : 0
-------------------------------------------------------------------
```

Starting back training as I don't think we achieved convergence:
```
-------------------------------------------------------------------
 iteration    simulation      bound        time (s)     solves  pid
-------------------------------------------------------------------
[...]

PR #779, opened 2024-09-04
Change BanditDuality to use a prior and softmax to randomize arms

odow:
x-ref https://github.com/odow/SDDP.jl/issues/777

@Thuener do you want to try this branch? (I haven't actually run it locally. Just edited without testing.)

Hopefully this PR:

 * Initializes with a default prior so that we do't have the `10` / constant bound issue
 * Randomizes the arm selection according to softmax, so to arms with similar scores will be randomly chosen, decaying quite quickly if the scores are differennt
 * Adds a `max(, 0.1)` to the time denominator to hedge against iterations that take a very short amount of time and increase the reward by too much.

### DIFF
--- a/src/plugins/duality_handlers.jl
+++ b/src/plugins/duality_handlers.jl
@@ -352,7 +352,7 @@ function BanditDuality()
     return BanditDuality(ContinuousConicDuality(), StrengthenedConicDuality())
 end
 
-function _choose_arm(handler::BanditDuality)
+function _update_arm(handler::BanditDuality)
     scores = map(handler.arms) do arm
         μ, σ = Statistics.mean(arm.rewards), Statistics.std(arm.rewards)
         # σ may be NaN if there are 0 or 1 observations, or if all observations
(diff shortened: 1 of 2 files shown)


Thuener:
Thanks!

I will try this one. 

One detail, as the bounds don't change in the middle of the backwards, it is better to change the arm just one time, no? That should be better for performance and if you change just for some nodes, it would break the cuts.

```julia
function prepare_backward_pass(
    node::Node,
    handler::BanditDuality,
    options::Options,
)
    if length(options.log) > handler.logs_seen
        _update_rewards(handler, options.log)
        handler.logs_seen = length(options.log)
        arm = _choose_best_arm(handler)
    end
    return prepare_backward_pass(node, handler.arms[handler.last_arm_index], options)
end
```

odow:
Good spotting. You might understand this better than I do. It's a while since I thought about it!

odow:
Merging because I think this is an improvement. 

My longer term project is to set up some sort of benchmarking so we could actually test if this was a good idea.

Issue #778, opened 2024-09-03
Performance Improvement on read/write cuts


Thuener:
When reading the cuts from file for a big model (with thousands of cuts) it can take couple of hours. My guess is because cut_selection is done every time we add a cut. Questions: 
- I was not able to find a way to disable cut_selection when reading the cuts. There is any way?
- We already did the cut selection when we first added the cuts; should we save only the cuts that we have already selected? This would reduce the cuts file size significantly and we would not have to do cut selection when reading the file.

Thuener:
I can confirm that removing the cut_selection will reduce the time to load the model from 50 minutes to less than 1 minute.

odow:
Adding a keyword to skip cut selection seems reasonable.

Thuener:
I'm trying to refactor the read and write function without changing the other functions... 

My idea was to have something like this:
```julia
function write_cuts_to_file2(
    model::SDDP.PolicyGraph{T},
    filename::String;
    node_name_parser::Function = string,
    save_only_selected_cuts = false,
) where {T}
    @info "Using new write_cuts"
    cuts = Dict{String,Any}[]
    for (node_name, node) in model.nodes
        if node.objective_state !== nothing || node.belief_state !== nothing
            error(
                "Unable to write cuts to file because model contains " *
                "objective states or belief states.",
            )
        end
        node_cuts = Dict(
            "node" => node_name_parser(node_name),
            "single_cuts" => Dict{String,Any}[],
            "multi_cuts" => Dict{String,Any}[],
            "risk_set_cuts" => Vector{Float64}[],
        )
        oracle = node.bellman_function.global_theta
        for (cut, state) in zip(oracle.cuts, oracle.sampled_states)
            @info "$(node_name), cut.constraint_ref $(cut.constraint_ref)"
            if !save_only_selected_cuts || cut.constraint_ref !== nothing
                intercept = cut.intercept
                for (key, π) in cut.coefficients
                    intercept += π * state.state[key]
                end
                push!(
                    node_cuts["single_cuts"],
                    Dict(
                        "intercept" => intercept,
                        "coefficients" => copy(cut.coefficients),
                        "state" => copy(state.state),
                    ),
                )
            end
        end
        for (i, theta) in enumerate(node.bellman_function.local_thetas)
            for (cut, state) in zip(theta.cuts, theta.sampled_states)
                if !save_only_selected_cuts || cut.constraint_ref !== nothing
                    intercept = cut.intercept
[...]

PR #781, opened 2024-09-05
Cut selection flags for write and read cuts functions

Thuener:
Related with #778, this PR adds: 
- cut_selection flag to use or not the cut selection algorithm when reading the files: This will significantly reduce the time to load a file
- write_only_selected_cuts to write only the selected cuts: This will reduce the file size and reading time significantly

### DIFF
--- a/src/plugins/bellman_functions.jl
+++ b/src/plugins/bellman_functions.jl
@@ -638,7 +638,7 @@ function write_cuts_to_file(
     model::PolicyGraph{T},
     filename::String;
     node_name_parser::Function = string,
-    write_only_selected_cuts::Bool = false
+    write_only_selected_cuts::Bool = false,
 ) where {T}
     cuts = Dict{String,Any}[]
     for (node_name, node) in model.nodes
--- a/src/plugins/bellman_functions.jl
+++ b/src/plugins/bellman_functions.jl
@@ -673,7 +673,9 @@ function write_cuts_to_file(
         end
         for (i, theta) in enumerate(node.bellman_function.local_thetas)
             for (cut, state) in zip(theta.cuts, theta.sampled_states)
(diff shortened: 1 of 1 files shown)


src/plugins/bellman_functions.jl
```diff
@@ -650,36 +656,40 @@ function write_cuts_to_file(
...
+                        Dict(
+                            "realization" => i,
+                            "intercept" => intercept,
+                            "coefficients" => copy(cut.coefficients),
+                            "state" => copy(state.state),
+                        ),
+                    )
+                end
+            end
+        end
```
odow, line 692: Why is this swapped around?

odow: Oh, it's just the weird diff

src/plugins/bellman_functions.jl
```diff
@@ -650,36 +656,40 @@ function write_cuts_to_file(
...
                         "intercept" => intercept,
                         "coefficients" => copy(cut.coefficients),
                         "state" => copy(state.state),
                     ),
                 )
             end
         end
+        for (i, theta) in enumerate(node.bellman_function.local_thetas)
+            for (cut, state) in zip(theta.cuts, theta.sampled_states)
+                if !write_only_selected_cuts || cut.constraint_ref !== nothing
```
odow, line 676: How about we keep the old indentation, and we do instead:
```suggestion
                if write_only_selected_cuts && cut.constraint_ref === nothing
                    continue
                end
```
I prefer fewer levels of indentation and early return/continue/break where possible.

odow (approved):
I'm happy if you're happy. You should have permission to merge.

We're missing a test, but I'm not too worried about that.

Thuener:
I can create the test. Any recommendation on the test?
Just write and read and calculate the bound is enough?

PR #789, opened 2024-10-13
Use disable_sigint to allow interrupts only between iterations

### LINKED ISSUE #767: Manually stopping training SIGINT Gurobi
I think this is an issue with Gurobi, but it is worth asking here... When using Gurobi I can't stop the training procedure using SIGINT event. Although I'm able to stop the train procedure the model will not work. For example, when I try to simulate the trained model I get the following error:
Gurobi Error 10017: Unable to set attribute 'LB'

odow:
Closes #767

### DIFF
--- a/src/plugins/parallel_schemes.jl
+++ b/src/plugins/parallel_schemes.jl
@@ -41,11 +41,17 @@ function master_loop(
 ) where {T}
     _initialize_solver(model; throw_error = false)
     while true
-        result = iteration(model, options)
-        options.post_iteration_callback(result)
-        log_iteration(options)
-        if result.has_converged
-            return result.status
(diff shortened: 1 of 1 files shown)


odow:
@Thuener this seemed to fix things locally.

src/plugins/parallel_schemes.jl
```diff
@@ -40,15 +40,23 @@ function master_loop(
...
+            options.post_iteration_callback(result)
+            log_iteration(options)
+            if result.has_converged
+                return result.status
+            end
+            return nothing
         end
     end
-    return
+    return return status
```
odow, line 59:
```suggestion
    return status
```

PR #790, opened 2024-10-14
Add a callback to modify the subproblem on numerical difficulty

### LINKED ISSUE #774: Add a callback to let the user modify a subproblem when numerical difficulties found
The default fallback should be 
https://github.com/odow/SDDP.jl/blob/d2495d6a9a0886048d560df7e29c7b220656594c/src/algorithm.jl#L303
and you should be able to do something differently.

odow:
Closes #774

### DIFF
--- a/src/algorithm.jl
+++ b/src/algorithm.jl
@@ -297,6 +297,58 @@ function _has_dual_solution(node::Node)
     return status in (JuMP.FEASIBLE_POINT, JuMP.NEARLY_FEASIBLE_POINT)
 end
 
+"""
+    set_numerical_difficulty_callback(
+        model::PolicyGraph,
+        callback::Function,
+    )
+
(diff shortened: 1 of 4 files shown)


src/algorithm.jl
```diff
@@ -292,19 +292,77 @@ function _has_primal_solution(node::Node)
...
+    if status == MOI.CONFLICT_FOUND
+        iis_model, _ = JuMP.copy_conflict(node.subproblem)
+        print(iis_model)
     end
-    if !_has_primal_solution(node)
-        model.ext[:numerical_issue] = true
+    return
+end
+SDDP.set_numerical_difficulty_callback(model, callback)
+```
```
odow, line 342: @joaquimg take a read of this docstring. Is it what you were after?

PR #797, opened 2024-10-23
Fix cut sharing in a graph with zero-probability arcs

### LINKED ISSUE #796: Cut sharing with Markov policy graph
I've got a SDDP model with a Markov chain used to define states and the transitions between them at each stage.

After noticing some strange behaviour in the simulated policy, the model was retrained with `refine_at_similar_nodes` set to `false`. This resolved the issue.

I've tracked the cause of the problem to there being 0 probabilities between some pairs of states at various stages. This seems to lead to SDDP not solving the model for those states, but the dual variable from the unsolved models are still being used to form cuts at other nodes in the policy graph.

I can't post a MFE here, but the above description will hopefully allow this issue to be reproduced.

odow:
Closes #796

@adow031 want to test this branch?

TODO
 - [x] add tests

### DIFF
--- a/src/algorithm.jl
+++ b/src/algorithm.jl
@@ -690,6 +690,11 @@ function backward_pass(
                 for other_index in options.similar_children[node_index]
                     copied_probability = similar(items.probability)
                     other_node = model[other_index]
+                    # Need to check that every child of other_node is in this
+                    # list
+                    other_children = Set(c.term for c in other_node.children)
+                    # Thhe order of setdiff(A, B) matters.
+                    @assert isempty(setdiff(other_children, Set(items.nodes)))
                     for (idx, child_index) in enumerate(items.nodes)
                         copied_probability[idx] =
                             get(options.Φ, (other_index, child_index), 0.0) *
--- a/src/algorithm.jl
+++ b/src/algorithm.jl
@@ -62,7 +62,7 @@ function get_same_children(model::PolicyGraph{T}) where {T}
     tmp = Dict{Set{T},Set{T}}()
     for (key, node) in model.nodes
         children =
-            Set(child.term for child in node.children if child.probabillity > 0)
+            Set(child.term for child in node.children if child.probability > 0)
         if length(children) == 0
             continue
         elseif haskey(tmp, children)
--- a/src/algorithm.jl
+++ b/src/algorithm.jl
@@ -61,7 +61,8 @@ end
 function get_same_children(model::PolicyGraph{T}) where {T}
     tmp = Dict{Set{T},Set{T}}()
     for (key, node) in model.nodes
-        children = Set(child.term for child in node.children)
+        children =
+            Set(child.term for child in node.children if child.probabillity > 0)
         if length(children) == 0
             continue
         elseif haskey(tmp, children)
--- a/src/algorithm.jl
+++ b/src/algorithm.jl
@@ -61,8 +61,7 @@ end
 function get_same_children(model::PolicyGraph{T}) where {T}
     tmp = Dict{Set{T},Set{T}}()
     for (key, node) in model.nodes
-        children =
-            Set(child.term for child in node.children if child.probability > 0)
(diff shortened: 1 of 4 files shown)


src/algorithm.jl
```diff
@@ -61,7 +61,8 @@ end
 function get_same_children(model::PolicyGraph{T}) where {T}
     tmp = Dict{Set{T},Set{T}}()
     for (key, node) in model.nodes
-        children = Set(child.term for child in node.children)
+        children =
+            Set(child.term for child in node.children if child.probabillity > 0)
```
odow, line 65:
```suggestion
            Set(child.term for child in node.children if child.probability > 0)
```

odow:
It'd be helpful if I could spell

adow031:
This issue appears to be resolved.

But looking at the fix: is the result that the there's no longer cut sharing when there are some 0 probabilities? I would prefer that the code just inefficiently solved children with 0 probabilities than didn't share cuts.

odow:
Currently we share cuts iff they have the identical set of children. We could improve that to nodes with have a subset of children. So yes, this might result in a less efficient solve; probably somewhere in between setting `refine_at_similar_nodes = false`, depending on the structure of the zeros.

odow:
But yeah, I'll take another go at fixing this.

odow:
Okay, I'm taking a look at this.

The issue is
https://github.com/odow/SDDP.jl/blob/f9216570ebd361b9e35b5678f2f5b83fe4da2222/src/algorithm.jl#L755-L757
which we added in https://github.com/odow/SDDP.jl/pull/224

The fix of just adding this is surprisingly subtle, because risk measures now need to deal with values that have zero probability mass

WorstCase is the obvious one, but it already checks this:
https://github.com/odow/SDDP.jl/blob/f9216570ebd361b9e35b5678f2f5b83fe4da2222/src/plugins/risk_measures.jl#L49-L57

test/visualization/value_functions.jl
```diff
@@ -135,11 +135,10 @@ function test_ValueFunction_belief_state()
     @test_throws AssertionError SDDP.evaluate(V11, Dict(:x => 1.0))
     b = Dict((1, 1) => 0.8, (1, 2) => 0.2)
     (y, duals) = SDDP.evaluate(V11, Dict(:x => 1.0); belief_state = b)
-    @test duals[:x] ≈ y ≈ 1.68
-
+    @test duals[:x] ≈ y ≈ 1.2
```
odow, line 138:
```suggestion
    @test duals[:x] ≈ y ≈ 1.68
```

test/visualization/value_functions.jl
```diff
@@ -135,11 +135,10 @@ function test_ValueFunction_belief_state()
...
     V12 = SDDP.ValueFunction(model[(1, 2)])
     (y, duals) = SDDP.evaluate(V12, Dict(:x => 1.0); belief_state = b)
-    @test duals[:x] ≈ y ≈ 1.68
+    @test duals[:x] ≈ y ≈ 1.2
```
odow, line 141:
```suggestion
    @test duals[:x] ≈ y ≈ 1.68
```

src/user_interface.jl
```diff
@@ -478,16 +478,8 @@ function MarkovianGraph(transition_matrices::Vector{Matrix{Float64}})
         for markov_state in 1:size(transition, 2)
             for last_markov_state in 1:size(transition, 1)
                 probability = transition[last_markov_state, markov_state]
-                if 0.0 < probability <= 1.0
```
odow, line 481: So @adow031: should we add zero-probability arcs for Markov transition matrices or keep the graph sparse?

odow:
Bump @adow031 interested in your thoughts

src/user_interface.jl
odow: I think the answer is that we want to keep them to enable cut re-use between the Markov nodes.

PR #806, opened 2024-11-27
Fix thread safety of RegularizedForwardPass

odow:
Closes https://github.com/odow/SDDP.jl/issues/791

PR #808, opened 2024-11-27
Fix thread safety of AlternativeForwardPass

odow:
Follow-up to #806


src/alternative_forward.jl
```diff
@@ -45,20 +46,24 @@ function forward_pass(
...
 """
 struct AlternativePostIterationCallback{T}
     model::PolicyGraph{T}
 end
 
 function (callback::AlternativePostIterationCallback)(result::IterationResult)
-    slave_update(callback.model, result)
+    # Only one thread is allowed to update the callback model at a time.
+    callback.lock() do
+        slave_update(callback.model, result)
```
odow, line 66:
```suggestion
        slave_update(callback.model, result)
        return
```

PR #819, opened 2025-01-20
Add variable bounds to incoming states in LagrangianDuality

### LINKED ISSUE #818: Domain of incoming states in the Lagrangian dual
I was reading @ChrisFuelOR's paper: https://optimization-online.org/wp-content/uploads/2024/08/Manuscript_Fuellner_Optimization_Online.pdf

Here's the most interesting section:

Image

Image

## SDDP.jl

When we dualize the incoming state bounds, we do not add new bonds:

https://github.com/odow/SDDP.jl/blob/aec55e2279c4abbd2a15ab1a430f6bb5a526976e/src/plugins/duality_handlers.jl#L178-L183

Our notation differs, but this is equivalent to the last choice of $\mathbb{R}$. I hadn't really thought about it before, but we can do much better if we restrict the domain to the union of the variable bounds in the previous nodes.

There's two subtleties that we need to account for:

 1. There might be many parent nodes. We need to take the union over all of them. Engineering problem but not an issue.
 2. The bounds might depend on the nodewise random variable. This one is surprisingly complicated. We could enumerate the in-sample realizations and then take the union. But this might exclude possible out-of-sample realizations. However, does that really matter? One option would be to bail back to the status quo if we detect the bounds changing. No harm no foul. The other option is to forge ahead. In theory, if, on a future backwards pass, we used a different set of samples, the cuts we had already generated might be invalid (and cut off parts of the cost-to-go function). But maybe that's okay.

odow:
Closes #818

### DIFF
--- a/docs/src/guides/add_multidimensional_noise.md
+++ b/docs/src/guides/add_multidimensional_noise.md
@@ -26,11 +26,12 @@ julia> model = SDDP.LinearPolicyGraph(
            SDDP.parameterize(subproblem, Ω, P) do ω
                JuMP.fix(x.out, ω.value)
                @stageobjective(subproblem, ω.coefficient * x.out)
-               println("ω is: ", ω)
            end
        end;
 
-julia> SDDP.simulate(model, 1);
+julia> for s in SDDP.simulate(model, 1)[1]
+           println("ω is: ", s[:noise_term_term])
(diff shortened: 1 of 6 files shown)


odow:
I still need to add possible integrality restrictions.

Issue #820, opened 2025-01-21
Interpolating between nodes in the graph


odow:
> Hello Oscar, I'd like to open this issue to ask about the possibility of interpolating the cut for an out-of-sample price point instead of using the cut from the closest price node as in Gjelsvik et al., 2010. Is this interpolation possible in SDDP.jl?
> Thank you for your time. 

 _Originally posted by @Remmy195 in [#707](https://github.com/odow/SDDP.jl/issues/707#issuecomment-2605453341)_

Remmy195:
I've looked at the objective state but have not implemented it because my price process is nonparametric so I have generated scenarios outside the model. Also setting the required parameters for the objective state (especially the lipschitz constant) is something I haven't been able to define yet.

I think using more nodes in the policy graph will perform well for the out-of-sample simulation; just at the expense of more computation.

Thank you!

odow:
> I think using more nodes in the policy graph will perform well for the out-of-sample simulation; just at the expense of more computation.

:+1:

Re computation time: we now support multi-threading, which is particularly effective with graphs with many nodes:
https://sddp.dev/stable/apireference/#SDDP.Threaded

Remmy195:
Ha nice! regarding parallelization; I've tried SDDP.Asynchronous() in the past but didn't get good solutions. I'll try SDDP.Threaded().

odow:
Yeah I should remove mention of `SDDP.Asynchronous` from the docs. No one should use it.

odow:
I'm tempted to close this as "won't-fix". I don't think I plan to add interpolation between nodes.

I've removed mention of Asynchronous from the docs and replaced it with Threaded in #821: https://sddp.dev/dev/guides/improve_computational_performance/#Parallelism

Remmy195:
I just want to note that I tested SDDP.Threaded(), and it was much faster with good solution. Thanks

odow:
> and it was much faster with good solution. Thanks

No problem :smile:

SolidAhmad:
> Yeah I should remove mention of `SDDP.Asynchronous` from the docs. No one should use it.

Hi Oscar, how come? Did the asynchronous scheme fail to show a tangible benefit? Does this mean SDDP.threaded() strictly uses the synchronous scheme?

odow:
`Threaded` is a different algorithm that uses multithreading instead of distributed parallelism. See https://jump.dev/JuMP.jl/stable/tutorials/algorithms/parallelism/

[...]

PR #821, opened 2025-01-21
[docs] document how to use the Threaded parallel scheme

odow:
x-ref https://github.com/odow/SDDP.jl/issues/820#issuecomment-2605927952

### DIFF
--- a/docs/src/guides/improve_computational_performance.md
+++ b/docs/src/guides/improve_computational_performance.md
@@ -66,7 +66,7 @@ Start Julia from a command line with `N` threads using the `--threads` flag:
 julia --threads N
 ```
 
-Thenn, pass an instance of [`SDDP.Threaded`](@ref) to the `parallel_scheme`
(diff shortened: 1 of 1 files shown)


docs/src/guides/improve_computational_performance.md
```diff
@@ -58,38 +58,16 @@ SDDP.train(model; cut_type = SDDP.MULTI_CUT)
...
-    @everywhere begin
-        import Pkg
-        Pkg.activate("/path/to/environment")
-    end
-    ```
-
-Currently SDDP.jl supports to parallel schemes, [`SDDP.Serial`](@ref) and
-[`SDDP.Asynchronous`](@ref). Instances of these parallel schemes should be passed to the
-`parallel_scheme` argument of [`SDDP.train`](@ref) and [`SDDP.simulate`](@ref).
+Thenn, pass an instance of [`SDDP.Threaded`](@ref) to the `parallel_scheme`
```
odow, line 69:
```suggestion
Then, pass an instance of [`SDDP.Threaded`](@ref) to the `parallel_scheme`
```

PR #830, opened 2025-02-27
Inner approximations

bfpc:
An initial version of inner approximations for Value functions.  Currently, supports only `LinearPolicyGraphs`, and no belief/objective states.


src/Inner.jl
```diff
@@ -0,0 +1,691 @@
+#  Copyright (c) 2017-25, Oscar Dowson and SDDP.jl contributors
+#  This Source Code Form is subject to the terms of the Mozilla Public
+#  License, v. 2.0. If a copy of the MPL was not distributed with this
+#  file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+module Inner
+# This code is heavily based on bellman_functions.jl
+
+using JuMP
```
odow, line 9:
```suggestion
import JuMP
```

Our suggested best practice is to use `import Package` within any other package. This forces you to prefix `Package.` on all calls, which is more explicit, and it avoids future breakage if JuMP starts exporting a new symbol. For example, if JuMP starts exporting `delete`, did we mean `Base.delete` or `JuMP.delete`?

bfpc: I guess I have now "converged" to a more reasonable PR.  I guess the next steps are:

- [ ] "duplicate" a graph, in order to build the "inner" policy graph from the "outer" one;
- [ ] Implement multi-vertices; do we (a) keep the notation `MULTI_CUT` (b) add new constants `SINGLE_VERTEX` & `MULTI_VERTEX` (c) replace by something like `NODEWISE` & `BRANCHWISE` (which could work for both and at the same time is not very common)
- [ ] Test for cyclic graphs / Markovian graphs / ...

but feel free to suggest other directions ;-)

odow: I think the next step is to add more tests.

See https://app.codecov.io/gh/odow/SDDP.jl/pull/830/blob/src/Inner.jl?dropdown=coverage

We should add a test for every red uncovered line.

There really a few that are unreachable, because there's an `error("Not implemented")` followed by some code. We should delete that code for now, and add a case that tests the error is hit.

odow: There's also a bunch of stuff for the risk-set and multi-cuts. Do you use any of that? I don't think so at the moment. So delete. 

I don't know if we _need_ multi-vertex support?

bfpc: Hi Oscar,

So, my idea would be that in the (near-ish) future we could provide something like this to evaluate the gap:

```julia
model = ...
SDDP.train(model)
SDDP.build_inner_bound(model)
```

Naturally, this would need a function to "copy" a `PolicyGraph` into a blank one, then perform a backward inner pass (or other strategy, but let's start with that).  As far as I see it, the next "low-hanging fruits" should be multi-cuts and (non-cyclic) Markovian Policy Graphs.

My idea is to create a "tracking issue" for those two (which seem independent to me), and maybe also add other features of `SDDP.jl` on a "wish-list".  My not-so-low next candidate would be Unicyclic policy graphs (because they are useful) but the backward inner pass should be modified to perform at least *some* loops in order to sufficiently approximate the inner functions.

docs/src/examples/inner_hydro_1d.jl
```diff
@@ -0,0 +1,114 @@
+#  Copyright (c) 2017-25, Oscar Dowson and SDDP.jl contributors.        #src
+#  This Source Code Form is subject to the terms of the Mozilla Public  #src
+#  License, v. 2.0. If a copy of the MPL was not distributed with this  #src
+#  file, You can obtain one at http://mozilla.org/MPL/2.0/.             #src
+
+# # Hydro 1d
+
+# This is a simple version of the hydro-thermal scheduling problem.
+# The goal is # to operate one hydro-dam and two thermal plants over time
```
odow, line 9:
```suggestion
# The goal is to operate one hydro-dam and two thermal plants over time
```

odow: You can see the preview here: https://sddp.dev/previews/PR830/examples/inner_hydro_1d/

docs/src/examples/inner_hydro_1d.jl
```diff
@@ -0,0 +1,114 @@
+#  Copyright (c) 2017-25, Oscar Dowson and SDDP.jl contributors.        #src
+#  This Source Code Form is subject to the terms of the Mozilla Public  #src
+#  License, v. 2.0. If a copy of the MPL was not distributed with this  #src
+#  file, You can obtain one at http://mozilla.org/MPL/2.0/.             #src
+
+# # Hydro 1d
```
odow, line 6: Let's give this a better name

docs/src/examples/inner_hydro_1d.jl
```diff
@@ -0,0 +1,114 @@
...
+
```
odow, line 7: Could we add something like
```
!!! warning
    The code in this example is experimental and the API may change in any future release.
```

odow:
I'll clean this up

PR #831, opened 2025-03-09
Improving user interfaces and test coverage in fpc/inner

rjmalves:
Contains some small improvements to the user interface while using the inner approximations. In general, the `inner_dp` function content was divided in smaller pieces:

- Adds an `InnerPolicyGraph` constructor that works in a similar way to the `SDDP.PolicyGraph`. By using it, the user does not need to explicitly build the `InnerBellmanFunction` object.
- Adds an `read_vertices_from_policy_graph` function that performes the single dynamic programming iteration from `inner_dp`, but receiving an object obtained from the `InnerPolicyGraph` constructor, without calculating bounds at the same time
- Adds some validations for arguments in these function calls

Also, the test coverage should be slightly improved with respect to the previous patch.

### DIFF
--- a/src/Inner.jl
+++ b/src/Inner.jl
@@ -74,18 +74,11 @@ end
 # TODO(bfpc): implement Objective and Belief states
 function _add_vertex_var_to_model(V::InnerConvexApproximation, vertex::Vertex)
     model = JuMP.owner_model(V.theta)
-    yᵀμ = JuMP.AffExpr(0.0)
     if V.objective_states !== nothing
         error("Objective states not yet implemented.")
-        for (y, μ) in zip(cut.obj_y, V.objective_states)
-            JuMP.add_to_expression!(yᵀμ, y, μ)
-        end
     end
     if V.belief_states !== nothing
         error("Belief states not yet implemented.")
-        for (k, μ) in V.belief_states
-            JuMP.add_to_expression!(yᵀμ, cut.belief_y[k], μ)
-        end
     end
 
     # Add a new variable to the convex combination constraints
--- a/src/Inner.jl
+++ b/src/Inner.jl
@@ -464,18 +464,36 @@ function model_type(model::SDDP.PolicyGraph{T}) where {T}
     return T
 end
 
+"""
+    _validate_linear_graph(graph::SDDP.Graph)
+
+Validates a given graph to be linear by counting the number of children
+at each node. This function does not replace the graph validation step from
+the main SDDP module and is meant to be used in conjunction with it while
+the Inner module does not support other graphs.
 
 """
-    _default_lipschitz_estimate(builder::Function)
-
-Makes an estimate of the lipschitz for using in the InnerBellmanFunction
-by calling the suproblem builder for the first node of the graph and taking
-the largest coefficient in the objective function.
-
-"""
-# TODO(rjmalves)
-function _default_lipschitz_estimate(builder::Function)
-    return 0.0
+function _validate_linear_graph(graph::SDDP.Graph)
+    node_count = length(graph.nodes)
+    has_leaf = false
+    for (node, children) in graph.nodes
+        children_count = length(children)
+        if children_count == 0
+            has_leaf = true
+        elseif children_count != 1
+            error(
+                "The graph is not linear since node $(node) has " *
+                "$(children_count) children and should have 1.",
+            )
+            
+        end
+    end
+    if !has_leaf
+        error(
+            "The graph is not linear since no leaf node was " *
+            "found among $(node_count) nodes.",
+        )
+    end
 end
 
 
@@ -552,9 +570,12 @@ function InnerPolicyGraph(
     # instead of simply calling LinearPolicyGraph, what should simplify things,
     # but will need to change when support to cyclic graphs is add.
     
+    # While only LinearGraphs are supported
+    _validate_linear_graph(graph)
+        
     # Cannot call PolicyGraph from the SDDP default interface since the
     # bellman_function named arg is marked as deprecated
-
+    
     # Spend a one-off cost validating the graph.
     SDDP._validate_graph(graph)
     # Construct a basic policy graph. We will add to it in the remainder of this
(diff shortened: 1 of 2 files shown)


src/Inner.jl
```diff
@@ -10,6 +10,8 @@ import JuMP
 import JSON
 import ..SDDP
 
+const MOI = SDDP.MOI
```
odow, line 13:
```suggestion
import JuMP: MOI
```

src/Inner.jl
```diff
@@ -469,6 +464,343 @@ function model_type(model::SDDP.PolicyGraph{T}) where {T}
...
+    for (node, children) in graph.nodes
+        children_count = length(children)
+        if children_count == 0
+            has_leaf = true
+        elseif children_count != 1
+            error(
+                "The graph is not linear since node $(node) has " *
+                "$(children_count) children and should have 1.",
+            )
+
```
odow, line 488:
```suggestion
```

src/Inner.jl
```diff
@@ -469,6 +464,343 @@ function model_type(model::SDDP.PolicyGraph{T}) where {T}
...
+            )
+            
+        end
+    end
+    if !has_leaf
+        error(
+            "The graph is not linear since no leaf node was " *
+            "found among $(node_count) nodes.",
+        )
+    end
```
odow, line 496:
```suggestion
    end
    return
```

src/Inner.jl
```diff
@@ -469,6 +464,343 @@ function model_type(model::SDDP.PolicyGraph{T}) where {T}
...
+end
+
```
odow, line 498:
```suggestion
```

src/Inner.jl
```diff
@@ -469,6 +464,343 @@ function model_type(model::SDDP.PolicyGraph{T}) where {T}
...
+    # This argument is the only hard difference from the "Cut-Based" PolicyGraph
+    lipschitz_constant = nothing,
+    # These arguments are deprecated
+    bellman_function = nothing,
+    direct_mode::Bool = false,
+) where {T}
+    # This function conveniently reuses code from the general PolicyGraph builder
+    # instead of simply calling LinearPolicyGraph, what should simplify things,
+    # but will need to change when support to cyclic graphs is add.
+
```
odow, line 572:
```suggestion
```

There are a bunch of other related formatting issues: https://github.com/odow/SDDP.jl/actions/runs/13751267942/job/38457363685?pr=831

test/Inner.jl
```diff
@@ -159,6 +159,137 @@ function test_Read_write_cuts_to_file()
...
+    graph = SDDP.LinearGraph(nstages)
+    cut_model = _create_model(graph)
+    SDDP.train(cut_model; iteration_limit = 50, print_level = 0)
+    vertex_model = create_policy_graph_with_inner_approximation()
+    SDDP.Inner.read_vertices_from_policy_graph(
+        vertex_model,
+        cut_model;
+        optimizer = HiGHS.Optimizer,
+    )
+    @test SDDP.calculate_bound(vertex_model) ≈ 45.833 atol = 0.1
```
odow, line 290: This failed on Julia 1.6. Is that expected?

test/Inner.jl
```diff
@@ -159,6 +159,137 @@ function test_Read_write_cuts_to_file()
     return
 end
 
+function create_policy_graph_with_inner_approximation()
+    nstages = 4
+    graph = SDDP.LinearGraph(nstages)
+
```
odow, line 165:
```suggestion
```
Please remove any unnecessary blank lines

test/Inner.jl
```diff
@@ -159,6 +159,137 @@ function test_Read_write_cuts_to_file()
...
+    # mising lower_bound even when sense = :Max
+    @test_throws Exception SDDP.Inner.InnerPolicyGraph(
+        build,
+        graph;
+        sense = :Max,
+        upper_bound = 1000.0,
+        optimizer = HiGHS.Optimizer,
+        lipschitz_constant = 10.0,
+    )
```
odow, line 232:
```suggestion
    )
    return
```

End every function with an explicit `return`

src/Inner.jl
rjmalves: I have configured the formatting extension here to prevent this from happening in future commits. Hope it works now, tested with the same code used by the workflow

test/Inner.jl
rjmalves: That is unexpected. It passes in the 1.0 workflow and here locally in 1.11... will investigate further and update here

rjmalves:
The formatting issues should be fixed for now. I will look deeper into the failing test in julia 1.6, which is unexpected

rjmalves:
It took some time to figure out the reason for failing only in julia 1.6, but after some time, we have something:

When the current implementation of the InnerBellmanFunction is used in SDDP.train, the first iterations might create some "ill-positioned" vertices that break the future improvement of the policy. I recall that @bfpc already had some insights about this happening in some cases.

For instance, in hydrothermal dispatch, it behaves like placing two vertices such that the line segment between them is greater then all the terms of the cost vector (including deficit), and these vertices cover a significant portion of the state space domain.

By comparing the forward noise samples in julia 1.6 and 1.11, which are different event with the same seed (and its reasonable), I've found that the specific 12345 default seed, with the test problem for the Inner submodule, causes a situation similar to the described above.

In order the test to stop failing, I've set another seed inside the test, but a more definitive solution is needed. This does not happen in other applications of the InnerBellmanFunction, such as just evaluating the (ex. upper) bound for a cut-based problem, or just making a simulation where the Bellman term is made of vertices instead of cuts. 

@odow if the formatting issues are solved, with this test passing, is it worth to merge the changes to this branch, or should we find a definitive solution to this issue before?

bfpc:
> It took some time to figure out the reason for failing only in julia 1.6, but after some time, we have something:

After talking with @rjmalves , there is no need for `SDDP.train()` to close the gap for `InnerBellmanFunction`s.  Actually, quite the opposite is expected, and the updated tests show it.

PR #834, opened 2025-03-27
Keep Markovian structure when parsing lattice from MSPFormat

ChrisFuelOR:
Added functionality in MSPFormat.jl to retain the Markovian structure of a lattice that is parsed using the _parse_lattice function.


odow:
The test failures are unrelated. I'll take a look

PR #836, opened 2025-03-31
Fix docs by pinning HiGHS_jll to v1.9.0

odow:


odow:
Temporary fix. I'll debug the underlying issue before the next release.

Issue #838, opened 2025-04-01
Tracking upstream bug in HiGHS


odow:
We're pinning the version of HiGHS_jll:

https://github.com/odow/SDDP.jl/blob/bf5abe7125c24c0212dd5b78bf4a6557bbe2d652/test/Project.toml#L16

to work around this issue in HiGHS:

https://github.com/ERGO-Code/HiGHS/issues/2267

odow:
Fixed upstream. Waiting for new release

PR #839, opened 2025-04-02
Free HiGHS_jll and work-around bug

odow:
x-ref #838


docs/src/examples/agriculture_mccardle_farm.jl
```diff
@@ -132,7 +132,9 @@ function test_mccardle_farm_model()
         end
         return
     end
-    SDDP.train(model)
+    # We set `cut_deletion_minimum` to work-around a bug in HiGHS_jll@1.10.0
```
odow, line 135:
```suggestion
    ## We set `cut_deletion_minimum` to work-around a bug in HiGHS_jll@1.10.0
```

docs/src/examples/agriculture_mccardle_farm.jl
```diff
@@ -132,7 +132,9 @@ function test_mccardle_farm_model()
...
+    # When fixed, revert to `SDDP.train(model)`
```
odow, line 136:
```suggestion
    ## When fixed, revert to `SDDP.train(model)`
```

PR #843, opened 2025-04-09
Prices are for different deficit "depths"

bfpc:
It is piecewise on the proportion of deficit on each subsystem, not with different prices per subsystem.  I didn't verify the original msppy python code.

### DIFF
--- a/papers/msppy/msppy_hydro_thermal.jl
+++ b/papers/msppy/msppy_hydro_thermal.jl
@@ -95,7 +95,7 @@ function msppy_hydro_thermal()
             inflow[i = 1:4] == inflow_initial[i]
         end)
         @stageobjective(sp,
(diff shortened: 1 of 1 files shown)


papers/msppy/msppy_hydro_thermal.jl
```diff
@@ -95,7 +95,7 @@ function msppy_hydro_thermal()
             inflow[i = 1:4] == inflow_initial[i]
         end)
         @stageobjective(sp,
-            sum(deficit_obj[i] * sum(deficit[i, :]) for i in 1:4) +
+            sum(deficit_obj[j] * sum(deficit[:, j]) for j in 1:4) +
```
odow, line 98: Good catch. Looks like I got confused updating 

https://github.com/lingquant/msppy/blob/dc85a2e8fa5243b3d5096d59085d9caad3ff2ede/examples/hydro_thermal/julia/test.jl#L94-L102

PR #848, opened 2025-04-16
Add ImportanceSamplingForwardPass

odow:
```julia
using SDDP
import HiGHS
import Random

model = SDDP.LinearPolicyGraph(;
    stages = 3,
    sense = :Min,
    lower_bound = 0.0,
    optimizer = HiGHS.Optimizer,
) do sp, t
    @variable(sp, 5 <= x <= 15, SDDP.State, initial_value = 10)
    @variable(sp, g_t >= 0)
    @variable(sp, g_h >= 0)
    @variable(sp, s >= 0)
    @constraint(sp, balance, x.out - x.in + g_h + s == 0)
    @constraint(sp, demand, g_h + g_t == 0)
    @stageobjective(sp, s + t * g_t)
    SDDP.parameterize(sp, [(0.0, 7.5), (3.0, 5.0), (10.0, 2.5)]) do w
        set_normalized_rhs(balance, w[1])
        set_normalized_rhs(demand, w[2])
        return
    end
    return
end

Random.seed!(1234)

SDDP.train(
    model;
    log_every_iteration = true,
    iteration_limit = 10,
    risk_measure = SDDP.Entropic(1e-1),
    forward_pass = SDDP.ImportanceSamplingForwardPass(),
    cut_type = SDDP.MULTI_CUT,
)
```

### DIFF
--- a/src/plugins/forward_passes.jl
+++ b/src/plugins/forward_passes.jl
@@ -398,3 +398,79 @@ function forward_pass(
     end
     return pass
 end
+
+struct ImportanceSamplingForwardPass <: AbstractForwardPass end
+
+function forward_pass(
+    model::PolicyGraph{T},
+    options::Options,
+    pass::ImportanceSamplingForwardPass,
+) where {T}
+    @assert isempty(model.belief_partition)
+    scenario_path = Tuple{T,Any}[]
+    sampled_states = Dict{Symbol,Float64}[]
+    cumulative_value = 0.0
+    incoming_state_value = copy(options.initial_state)
+    node_index = sample_noise(model.root_children)
+    node = model[node_index]
+    noise = sample_noise(node.noise_terms)
+    while node_index !== nothing
+        node = model[node_index]
+        lock(node.lock)
+        try
+            push!(scenario_path, (node_index, noise))
+            subproblem_results = solve_subproblem(
+                model,
+                node,
+                incoming_state_value,
+                noise,
+                scenario_path;
+                duality_handler = nothing,
+            )
+            cumulative_value += subproblem_results.stage_objective
+            incoming_state_value = copy(subproblem_results.state)
+            push!(sampled_states, incoming_state_value)
+            if isempty(node.bellman_function.local_thetas)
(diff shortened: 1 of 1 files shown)


joaquimg:
This is it:
https://arxiv.org/abs/2307.13190
?

odow:
Ah I knew there was something! A student of Dave is trying to do something similar.

odow:
💯 This was the next step I had in mind:

image

They're coming at it from a different direction/motivation, but it's the same core technique.

src/plugins/forward_passes.jl
```diff
@@ -398,3 +398,79 @@ function forward_pass(
...
+                _ = SDDP.adjust_probability(
+                    options.risk_measures[node_index],
+                    adjusted_probability,
+                    nominal_probability,
+                    support,
+                    objectives,
+                    model.objective_sense == MOI.MIN_SENSE,
+                )
+                terms = SDDP.Noise.(support, adjusted_probability)
+                node_index, noise = SDDP.sample_noise(terms)
```
odow, line 463: @Danjiang2000 it's a complete implementation. The two for-loops build the nominal probability pmf, then we use `SDDP.adjust_probability` to get the risk-adjusted probabilities, and then we sample a new node and noise from that distribution.

odow:
Yeah I thought of this. I initially considered looping through the cuts evaluating them before I realized I could reach in and find the right `JuMP.value(theta)`.

I guess the point is that there is a lot of information in the disaggregated cuts that we are throwing away just for the sake of adding fewer constraints.

src/plugins/forward_passes.jl
odow: > Does this sound like a good plan if I want to implement a complete importance sampling experiment?

Yes, I think at least for now.

What this is really asking for the ability to sample with the risk-adjusted probabilities in `SDDP.simulate` instead of `SDDP.train`. We either need to write something separate for `SDDP.simulate`, or change `SDDP.simulate` to use the `AbstractForwardPass` machinery.

odow:
There are still some issues with this, but it's a useful start

PR #856, opened 2025-06-11
Add FixedDiscreteDuality

odow:
Motivated by this forum discussion: https://or.stackexchange.com/questions/13191/benders-cuts-discrete-variables

This was actually the technique Tony and I used for the original MilkOutputOptimizer model in 2014.

Edit: I plucked up the courage to upload the report: https://github.com/odow/SDDP.jl/blob/master/papers/other/dowson_2014.pdf

The relevant part of the report _massively_ glosses over the details (like the fact the cuts might be invalid):

image

Historical tidbit: the first version of SDDP.jl used gurobipy...

image

### DIFF
--- a/docs/src/apireference.md
+++ b/docs/src/apireference.md
@@ -347,6 +347,11 @@ SDDP.StrengthenedConicDuality
 SDDP.BanditDuality
 ```
 
+## `FixedDiscreteDuality`
+```@docs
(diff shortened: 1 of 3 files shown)


odow:
See https://github.com/odow/SDDP.jl/pull/865

PR #888, opened 2026-02-27
[docs] rename tutorials to more explicit "How to"

odow:


docs/src/guides/deterministic_equivalent.md
```diff
@@ -0,0 +1,51 @@
+## How to solve the deterministic equivalent
```
odow:
```suggestion
# How to solve the deterministic equivalent
```

PR #903, opened 2026-07-27
Add SDDP.plot(::PolicyGraph) for visualizing the graph layout

### LINKED ISSUE #896: Add a function to plot the graph
Because it's pretty, and because it might help communicate the models that people are building.

odow:
Closes #896


odow:
Nice nice nice. This is working really well. https://sddp.dev/previews/PR903/tutorial/capacity_expansion/

odow:
This is such a big win. You can argue about some of the layout decisions, but layout is hard. I'd prefer to just use the library than code something ourselves.

image

image

image

image

PR #917, opened 2026-09-05
ci: update checkout, julia cache, and codecov to current majors

katsugtgz:
Updates workflow actions to current majors. The latest CI run on master logs a Node.js 20 deprecation warning for checkout@v4, codecov-action@v4, and julia-actions/cache@v1.

Changes:
- actions/checkout v4 -> v7 in ci.yml, doc_cleanup.yml, documentation.yml, format_check.yml
- julia-actions/cache v1 -> v3 in ci.yml. Note: since v2 the cache is also saved when the job fails. Set `save-always: false` if you want the old behavior.
- codecov/codecov-action v4 -> v7 in ci.yml, with `file:` renamed to `files:` per the v5 migration guide. The CODECOV_TOKEN input is unchanged.
- julia-actions/buildpkg/runtest/processcoverage stay on v1, which is still their current major.

Validation:
- actionlint 1.7.12 passes on the four updated workflows
- checkout v7 credential change is compatible with the git push in doc_cleanup.yml (credentials are persisted in a file under $RUNNER_TEMP; git commands keep working)

Scope: .github/workflows/ only.

Note: this PR comes from a fork. If workflow runs show action_required, they need a maintainer approval click; I will not retry-push to force it.

Commit author katsugtgz, unsigned.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,11 @@
+# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/" # Location of package manifests
+    schedule:
+      interval: "weekly"
+  - package-ecosystem: "julia"
+    directory: "/" # Location of package manifests
+    schedule:
+      interval: "weekly"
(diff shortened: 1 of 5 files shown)


odow:
Thanks!

PR #925, opened 2026-09-11
[docs] update capacity expansion tutorial to invest in wind

odow:


odow:
A note to myself for when I pick this back up:

 * For some reason, we add cuts with large coefficients
 * HiGHS will not accept 10^15 or larger
 * This could be a bug in the model, or just an artefact of the dual solution at the edge of the feasible region (when reservoir is full, it probably supports any dual > 0)
 * It could also be a bug in HiGHS simplex/presolve

